PR #34, opened 2019-06-18
added more devs to auto assign bot

purujitb:


purujitb:
@vbvg2008 we should wait for the CI part to complete before approving ;)

vbvg2008:
how come CI takes so long for a branch that has no test at all

vbvg2008:
oops my bad, it's r0.2

PR #35, opened 2019-06-19
First pass of saliency maps

TortoiseHam:
Adds saliency map functionality and moves both saliency mapping and log visualization into a new 'visualization' module which can be invoked from the main CLI


setup.py
```diff
@@ -37,10 +35,11 @@
           'scipy',
           'pytest',
           'pytest-cov',
-          'opencv-python'
+          'opencv-python',
+          'tensorflow'
```
vbvg2008: it is not recommended to put tensorflow in the setup.py.  because users may want to use tensorflow-gpu instead

TortoiseHam: ah, ok

PR #66, opened 2019-07-15
Adversarial and Mixup showcases

TortoiseHam:
New showcase examples on the CIFAR10 dataset

remove time axis from training (+1 squashed commit)
Squashed commits:
[c64ea68] time dimension (+8 squashed commits)
Squashed commits:
[61d4f71] cleanup
[69f3fc4] improve mixup efficiency
[56c3cfb] cleanup
[7387f58] disable recursive search by default for log visualization
[3bf7a77] remove dataset hack
[5efb29b] back to relu
[c4f63f1] rule around 2k example/sec, swish around 1.3k
[475167c] first pass: 95 examples / sec

### DIFF
--- a/fastestimator/architecture/lenet.py
+++ b/fastestimator/architecture/lenet.py
@@ -1,7 +1,8 @@
-from tensorflow.keras import layers, models
+from tensorflow.python.keras import layers, Sequential
 
-def LeNet(input_shape=(28,28,1), classes=10):
-    model = models.Sequential()
+
+def LeNet(input_shape=(28, 28, 1), classes=10):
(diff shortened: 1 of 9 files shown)


fastestimator/architecture/lenet.py
```diff
@@ -1,7 +1,8 @@
-from tensorflow.keras import layers, models
+from tensorflow.python.keras import layers, Sequential
```
vbvg2008: I don't think we use this import

TortoiseHam: lines 5 and 6

PR #80, opened 2019-08-03
Add caricature visualization capability

TortoiseHam:
An interpretability module as described here, adapted for TF2: https://github.com/tensorflow/lucid/issues/121

plantdog

add caricature into FE bash (+3 squashed commits)
Squashed commits:
[e6a10e3] cleaning
[b087313] merge cleanup
[abc32e1] more cleanup (+6 squashed commits)
Squashed commits:
[fb147c4] clean up / consolidate code
[da87600] fft space optimization is working
[d528cd3] add image smoothing to generate cleaner reconstructions
[a8dd0b3] fix static augmentation to work within gradient tapes / graph mode (seems there may be a bug with boolean tensors at the moment)
[4ff1fda] system is starting to work better - results still don't seem consistent with google publication though. augmentation module doesn't work inside gradient tape due to an issue with boolean member variables. still investigating.
[c6b3656] layer-based image generation. doesn't seem to be working the way it does in the papers though yet. theirs may be neuron or channel specific

### DIFF
--- a/bin/fastestimator
+++ b/bin/fastestimator
@@ -7,6 +7,7 @@ import sys
 from pyfiglet import Figlet
 
 from fastestimator.util.util import parse_cli_to_dictionary, get_gpu_count
+from fastestimator.visualization.caricature import load_and_caricature
 from fastestimator.visualization.layer_umap import umap_layers
 from fastestimator.visualization.parse_logs import parse_folder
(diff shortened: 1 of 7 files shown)


TortoiseHam:
Example usage:

fastestimator visualize caricature inceptionV3.h5 plants.png dogcat.png --dictionary imagenet_class_index.json --strip-alpha --layers 164

PR #88, opened 2019-08-06
merge the prediction and batch dictionary for Loss and Trace calculation

vbvg2008:
now users only need to worry about one dictionary


fastestimator/network/loss.py
```diff
@@ -19,13 +19,12 @@ class Loss:
     def __init__(self):
         pass
 
-    def calculate_loss(self, batch, prediction, mode):
+    def calculate_loss(self, batch, state):
         """this is the function that calculates the loss given the batch data
         
         Args:
             batch (dict): batch data before forward operation
```
TortoiseHam: this is now batch data after forward op right?

vbvg2008: yep

TortoiseHam:
should we define static strings somewhere for the state keys so that we don't have to keep typing the string everywhere?

showcase/image_classification/lenet_cifar10_mixup.py
```diff
@@ -24,28 +24,20 @@
 from fastestimator.pipeline.pipeline import Pipeline
 from fastestimator.pipeline.preprocess import Minmax
 
-
```
TortoiseHam: your IDE isn't following pep8 formatting standards - there should be 2 spaces here

showcase/image_detection/retinanet_svhn.py
```diff
@@ -35,15 +35,13 @@ def forward(self, data):
             data[idx] = np.array([int(x) for x in elem[1:-1].split(',')])
         return data
 
-
```
TortoiseHam: pep8 formatting violation again in this file

fastestimator/estimator/estimator.py
```diff
@@ -73,62 +74,66 @@ def _add_traces(self):
...
         for train_step, batch in enumerate(self.training_fn().take(self.steps_per_epoch * self.epochs)):
             if train_step % self.steps_per_epoch == 0:
                 self.epoch = train_step // self.steps_per_epoch
-                self._run_traces_on_epoch_begin(mode="train", logs={"epoch": self.epoch})
-            self._run_traces_on_batch_begin(mode="train", logs= {"epoch": self.epoch, "step": train_step, "size": self.pipeline.batch_size})
-            prediction, loss = self.forward_step(batch, mode="train", epoch=self.epoch)
-            self._run_traces_on_batch_end(mode="train", logs= {"epoch": self.epoch, "step": train_step, "size": self.pipeline.batch_size, "batch": batch, "prediction": prediction, "loss": loss})
+                self._run_traces_on_epoch_begin({"mode": "train", "epoch": self.epoch})
+            self._run_traces_on_batch_begin( {"mode": "train", "epoch": self.epoch, "step": train_step, "batch_size": self.pipeline.batch_size})
+            # prediction, loss = self.forward_step(batch, {"mode":"train", "epoch": self.epoch, "step": train_step})
```
TortoiseHam: why not pass epoch and step to the forward?

fastestimator/network/loss.py
TortoiseHam: (comment was to point out the doc string is wrong)

PR #94, opened 2019-08-12
Added Fast Style Transfer showcase

jphong89:
- Added style transfer net & template
- Added mscoco dataset api

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,10 @@
+[style]
+based_on_style = pep8
+column_limit = 120
+coalesce_brackets = True
+dedent_closing_brackets = False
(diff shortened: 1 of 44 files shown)


fastestimator/architecture/stnet.py
```diff
@@ -0,0 +1,125 @@
+import tensorflow as tf
+
+class ReflectionPadding2D(tf.keras.layers.Layer):
```
TortoiseHam: is this class generic enough that it would be useful to have as a separate file?

jphong89: I think this type of padding is specific to problems like image to image translation (e.g.,  cycleGAN) and style transfer.

PR #110, opened 2019-08-20
remove horovod related setup in cli, update setup, add fe version fla…

vbvg2008:
…g by fe.__version__

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,16 +1,24 @@
-from __future__ import absolute_import
+import os
+import re
+
+from setuptools import find_packages, setup
 
 
-from setuptools import setup, find_packages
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -1,16 +1,24 @@
-from __future__ import absolute_import
+import os
+import re
 
+from setuptools import find_packages, setup
 
-from setuptools import setup, find_packages
+
+def get_version():
```
TortoiseHam: this seems way more complicated than the old version number scheme. what's the benefit?

vbvg2008: being able to get fastestimator version by fe.__version__

PR #114, opened 2019-08-21
Traces are Ops

TortoiseHam:
clarifying comments (+1 squashed commit)
Squashed commits:
[14ba668] partial update of the GifGenerator trace (in line with op syntax, but may not actually work still since wasn't working before) (+5 squashed commits)
Squashed commits:
[decd333] update the other Traces
[1dc96af] update confusion matrix
[55ba969] fix the accuracy trace
[249a684] update logger traces to op-based
[7ab45ef] preliminary shuffling of functions

### DIFF
--- a/fastestimator/estimator/estimator.py
+++ b/fastestimator/estimator/estimator.py
@@ -17,7 +17,8 @@ from collections import ChainMap
 import numpy as np
 import tensorflow as tf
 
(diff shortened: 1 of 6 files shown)


vbvg2008:
this PR requires further discussion, please don't merge it yet

PR #115, opened 2019-08-21
Fixed dcgan mnist example

chm123:


showcase/image_generation/dcgan_mnist.py
```diff
@@ -88,18 +88,20 @@ def forward(self, data, state):
 
 def get_estimator():
```
TortoiseHam: we might as well make epoch and batch_size passable params here

PR #120, opened 2019-08-23
Added Wide and Deep Neural Network example

gaurav272333:


showcase/tabular/wide_and_deep.py
```diff
@@ -0,0 +1,145 @@
...
+from fastestimator.network.loss import BinaryCrossentropy
+from fastestimator.estimator.trace import Accuracy
+from fastestimator.dataset import movielens20m_data
+
+_USER_INPUT_DIM = 138493
+_MOVIE_INPUT_DIM = 26744
+_USER_EMBEDDING_DIM = 64
+_MOVIE_EMBEDDING_DIM = 15
+
+GENRES = [
```
TortoiseHam: the should probably be in the data file

fastestimator/dataset/movielens20m_data.py
```diff
@@ -0,0 +1,108 @@
...
+    # Creation of Train and Evaluation Sets
+    train_df = df.sample(frac=0.8, random_state=0)
+    eval_df = df.drop(train_df.index)
+
+    train_df = train_df.reset_index(drop=True)
+    eval_df = eval_df.reset_index(drop=True)
+
+    # Applying Standard Scaler
+    Std_cols = ["genres_count", "year", "month", "day"]
+    scaler = StandardScaler()
```
TortoiseHam: should we be demonstrating how to do this in Pipeline instead?

showcase/tabular/wide_and_deep.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # deep Part
+    x_deep1 = concatenate([embed_layer, input_genres, input_time])
+    x_deep2 = Dense(256, activation="relu")(x_deep1)
+    x_deep3 = Dense(128, activation="relu")(x_deep2)
+    x_deep4 = Dense(64, activation="relu")(x_deep3)
+
+    # wide part
+    x_wide1 = concatenate([input_genres, input_time])
+    x_wide2 = Dense(16, activation="relu")(x_wide1)
```
TortoiseHam: isn't this 'narrower' than x_deep1?

showcase/tabular/wide_and_deep.py
gaurav272333: Sure. This makes sense. I will move this to the data file.

fastestimator/dataset/movielens20m_data.py
gaurav272333: As this needs to be applied on entire dataset, we won't be able to apply this in Pipeline.

PR #138, opened 2019-08-30
Multi-gpu support

vbvg2008:
now FE can run on multi-gpu without code change

### DIFF
--- a/fastestimator/estimator/estimator.py
+++ b/fastestimator/estimator/estimator.py
@@ -40,7 +40,8 @@ class Estimator:
         self.log_steps = log_steps
         self.num_devices = get_num_devices()
         if self.num_devices > 1:
-            self.distribute_strategy = tf.distribute.MirroredStrategy()
+            # self.distribute_strategy = tf.distribute.MirroredStrategy()
+            self.distribute_strategy = None
         else:
             self.distribute_strategy = None
 
@@ -95,8 +96,8 @@ class Estimator:
                 batch = next(ds_iter)
                 global_batch_size = self.pipeline.get_global_batch_size(epoch)
                 state["batch_size"] = global_batch_size
-                ops, model_list, loss_list = self.network.load_epoch(epoch, mode)
-                self.network.run_step(batch, ops, model_list, loss_list, state, warm_up=True)
+                ops, model_list, losses_epoch = self.network.load_epoch(epoch, mode)
+                self.network.run_step(batch, ops, model_list, losses_epoch, state, warm_up=True)
 
     def train(self):
         train_start = time.perf_counter()
@@ -109,7 +110,7 @@ class Estimator:
                 max_steps = self.steps_per_epoch
             else:
                 max_steps = min(self.pipeline.num_examples["train"]) // global_batch_size
-            ops, model_list, loss_list = self.network.load_epoch(epoch, "train")
+            ops, model_list, losses_epoch = self.network.load_epoch(epoch, "train")
             self._run_traces_on_epoch_begin({"mode": "train", "epoch": epoch, "train_step": train_step})
             for batch_idx in range(max_steps):
                 batch = next(ds_iter)
@@ -124,7 +125,7 @@ class Estimator:
                 prediction = self.forward_step(batch,
                                                ops,
                                                model_list,
-                                               loss_list, {
+                                               losses_epoch, {
                                                    "mode": "train", "batch_size": global_batch_size
                                                })
                 batch = ChainMap(prediction, batch)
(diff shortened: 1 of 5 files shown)


fastestimator/estimator/estimator.py
```diff
@@ -140,23 +140,41 @@ def _warmup(self):
             for epoch in signature_epochs:
                 ds_iter = self.pipeline.dataset_schedule[mode].get_current_value(epoch)
                 batch = next(ds_iter)
-                global_batch_size = self.pipeline.get_global_batch_size(epoch)
+                if self.distribute_strategy and mode == "train":
```
TortoiseHam: do we still need to limit to train mode since we're doing the prediction concatenation now?

fastestimator/network/network.py
```diff
@@ -85,27 +87,29 @@ def _get_signature_epoch(self, mode):
...
         ops = self.op_schedule[mode].get_current_value(epoch)
         model_list = self.model_schedule[mode].get_current_value(epoch)
-        loss_list = []
+        losses_epoch = []
         for model in model_list:
-            if model.loss_name not in loss_list:
-                loss_list.append(model.loss_name)
-        self.loss_list = loss_list
-        return ops, model_list, loss_list
+            if model.loss_name not in losses_epoch:
```
TortoiseHam: we might want to make losses_epoch a set rather than list to speed up the 'not in' check

fastestimator/network/network.py
```diff
@@ -32,7 +34,7 @@ def __init__(self, ops):
         self.current_epoch_model = {}
         self.model = {}
         self.all_losses = []
-        self.loss_list = []
+        self.losses_epoch = []
```
TortoiseHam: this name might make more sense as epoch_losses to be more consistent with 'all_losses' from above

fastestimator/network/network.py
```diff
@@ -128,9 +133,11 @@ def _forward(batch, state, ops):
...
-        for loss_key in loss_list:
-            loss = batch[loss_key]
+    def _reduce_loss(batch, global_batch_size, losses_epoch, warm_up):
+        reduced_loss = {}
+        for loss_name in losses_epoch:
+            element_wise_loss = batch[loss_name]
             if warm_up:
-                assert loss.shape.num_elements() > 1, "please make sure loss is element-wise loss"
-            batch[loss_key] = tf.reduce_sum(loss) / global_batch_size
+                assert element_wise_loss.shape.num_elements() > 1, "please make sure loss is element-wise loss"
```
TortoiseHam: what if batch size == 1?

fastestimator/pipeline/pipeline.py
```diff
@@ -99,6 +100,8 @@ def prepare(self, distribute_strategy=None):
         for mode in self.mode_list:
             self._get_feature_name(mode)
             self._extract_dataset(mode)
+            if mode != "train":
+                distribute_strategy = None
```
TortoiseHam: as in earlier comment, can we not distribute the eval mode?

showcase/image_generation/dcgan_mnist.py
```diff
@@ -92,10 +92,10 @@ def forward(self, data, state):
         return data
 
 
-def get_estimator(batch_size=256, epochs=10):
+def get_estimator(batch_size=256, epochs=50):
     # prepare data
-    (x_train, _), (x_eval, _) = tf.keras.datasets.mnist.load_data()
-    data = {"train": {"x": np.expand_dims(x_train, -1)}, "eval": {"x": np.expand_dims(x_eval, -1)}}
+    (x_train, _), (_, _) = tf.keras.datasets.mnist.load_data()
```
TortoiseHam: if we're not doing eval do we want to add that data into the train set?

fastestimator/network/network.py
vbvg2008: fixed

fastestimator/network/network.py
vbvg2008: I think the computational complexity for member check is the same.   Also, set is not iterable, which will create problems when we try to iterate through them.

showcase/image_generation/dcgan_mnist.py
vbvg2008: I think we will do the same as tensorflow tutorial.   no need to add that.

fastestimator/network/network.py
TortoiseHam: sets are iterable. also, unless python did sets totally wrong, membership check in set is O(1) vs in list O(n)

fastestimator/pipeline/pipeline.py
vbvg2008: fixed

fastestimator/network/network.py
vbvg2008: fixed

PR #143, opened 2019-09-03
Trace BugFix

TortoiseHam:
bug was introduced since calling set('string') returns {'s', 't', 'r', 'i', 'n', 'g'} instead of the desired {'string'}. Also, the Trace sorting needs all output keys defined in the network graph, not just the loss keys.

### DIFF
--- a/fastestimator/estimator/estimator.py
+++ b/fastestimator/estimator/estimator.py
@@ -19,6 +19,7 @@ import tensorflow as tf
 
 from fastestimator.estimator.trace import Logger, MonitorLoss, Trace
 from fastestimator.util.cli_util import draw
+from fastestimator.util.schedule import Scheduler
 from fastestimator.util.util import get_num_devices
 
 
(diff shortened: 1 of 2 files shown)


fastestimator/estimator/estimator.py
```diff
@@ -90,7 +91,10 @@ def _sort_traces(self):
         sorted_traces = []
         available_outputs = {
             "num_devices", "mode", "epoch", "train_step", "batch_idx", "batch_size", "batch", "elapsed_time"
-        } | set(self.network.all_losses)
+        } | set().union(*[{item.outputs} if isinstance(item.outputs, str) else set(item.outputs)
+                          for sublist in (list(op.epoch_dict.values()) if isinstance(op, Scheduler) else [op]
+                                          for op in self.network.ops) for item in sublist])
+        # available_outputs starts with all of our default keys plus all of the outputs listed within any network ops
```
vbvg2008: so all these efforts is because we don't want to use list as member check?

TortoiseHam: lol, no, we don't have that info anywhere in list form either

PR #148, opened 2019-09-04
Add basic ModelCheckpoint class to trace

chm123:
Update modelckpt signature

Remove unused variable

Support model_names and save_freq

Complete ModelCheckpoint trace

Update showcases

### DIFF
--- a/fastestimator/estimator/estimator.py
+++ b/fastestimator/estimator/estimator.py
@@ -38,6 +38,8 @@ class Estimator:
         self.validation_steps = validation_steps
         self.traces = traces
         self.log_steps = log_steps
+        self.inputs = None
+        self.do_eval = None
         self.num_devices = get_num_devices()
         if self.num_devices > 1:
             self.distribute_strategy = tf.distribute.MirroredStrategy()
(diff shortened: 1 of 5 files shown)


fastestimator/estimator/trace.py
```diff
@@ -911,3 +931,125 @@ def get_logdir(self):
...
+            self._last_saved_files = []
+
+        # save all specified models
+        if (self.save_best_only and is_best_model) or (not self.save_best_only):
+            for model_key in self.model_names:
+                save_path = os.path.join(
+                    self.save_dir, '{}_epoch_{}_step_{}.h5'.format(model_key, state['epoch'], state['train_step']))
+                self.network.model[model_key].save(save_path, include_optimizer=False)
+                self._last_saved_files.append(save_path)
+                print("FastEstimator-ModelCheckpoint: Saving '{}' to {}".format(model_key, save_path))
```
jphong89: Can you add a verbose so that users have a choice to disable printing the message?

fastestimator/estimator/estimator.py
```diff
@@ -38,6 +38,8 @@ def __init__(self,
         self.validation_steps = validation_steps
         self.traces = traces
         self.log_steps = log_steps
+        self.inputs = None
+        self.do_eval = None
```
jphong89: do these changes break any other components?

chm123: pylint is complaining property defined outside __init__()

jphong89: why is do_eval changed from False to None? Wouldn't it affect other components of estimator?

PR #157, opened 2019-09-13
Introducing.... LRController

vbvg2008:
LRController is unifying lrscheduler and reducelronplateau.

### DIFF
--- a/apphub/image_classification/lenet_mnist.py
+++ b/apphub/image_classification/lenet_mnist.py
@@ -26,7 +26,7 @@ from fastestimator.network.model import FEModel, ModelOp
 from fastestimator.pipeline.processing import Minmax
 
 
-def get_estimator(epochs=2, batch_size=32):
+def get_estimator(epochs=2, batch_size=32, save_dir=tempfile.mkdtemp()):
     # step 1. prepare data
     (x_train, y_train), (x_eval, y_eval) = tf.keras.datasets.mnist.load_data()
     data = {
@@ -48,7 +48,7 @@ def get_estimator(epochs=2, batch_size=32):
     traces = [
         Accuracy(true_key="y", pred_key="y_pred", output_name='acc'),
         LRController(model_name="lenet", lr_schedule=CyclicLRSchedule()),
-        ModelSaver(model_name="lenet", save_dir=tempfile.mkdtemp(), save_best="acc", save_best_mode="max")
+        ModelSaver(model_name="lenet", save_dir=save_dir, save_best="acc", save_best_mode="max")
     ]
     estimator = fe.Estimator(network=network, pipeline=pipeline, epochs=epochs, traces=traces)
     return estimator
--- a/apphub/image_classification/lenet_mnist.py
+++ b/apphub/image_classification/lenet_mnist.py
@@ -49,5 +49,5 @@ def get_estimator(epochs=2, batch_size=32):
 
 
 if __name__ == "__main__":
-    estimator = get_estimator()
-    estimator.fit()
+    est = get_estimator()
+    est.fit()
--- a/apphub/image_classification/lenet_mnist.py
+++ b/apphub/image_classification/lenet_mnist.py
@@ -12,12 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import tempfile
+
 import numpy as np
 import tensorflow as tf
 
 import fastestimator as fe
 from fastestimator.architecture import LeNet
-from fastestimator.estimator.trace import Accuracy, LRController, ModelCheckpoint
+from fastestimator.estimator.trace import Accuracy, LRController, ModelSaver
 from fastestimator.network.loss import SparseCategoricalCrossentropy
 from fastestimator.network.lrschedule import CyclicLRSchedule
 from fastestimator.network.model import FEModel, ModelOp
@@ -36,14 +38,17 @@ def get_estimator(epochs=2, batch_size=32):
         }
     }
     pipeline = fe.Pipeline(batch_size=batch_size, data=data, ops=Minmax(inputs="x", outputs="x"))
+
     # step 2. prepare model
     model = FEModel(model_def=LeNet, model_name="lenet", optimizer="adam")
     network = fe.Network(
         ops=[ModelOp(inputs="x", model=model, outputs="y_pred"), SparseCategoricalCrossentropy(inputs=("y", "y_pred"))])
+
     # step 3.prepare estimator
     traces = [
         Accuracy(true_key="y", pred_key="y_pred", output_name='acc'),
-        LRController(model_name="lenet", lr_schedule=CyclicLRSchedule())
+        LRController(model_name="lenet", lr_schedule=CyclicLRSchedule()),
+        ModelSaver(model_name="lenet", save_dir=tempfile.mkdtemp(), save_best="acc", save_best_mode="max")
     ]
     estimator = fe.Estimator(network=network, pipeline=pipeline, epochs=epochs, traces=traces)
     return estimator
--- a/apphub/image_classification/lenet_mnist.py
+++ b/apphub/image_classification/lenet_mnist.py
@@ -17,8 +17,9 @@ import tensorflow as tf
 
 import fastestimator as fe
 from fastestimator.architecture import LeNet
-from fastestimator.estimator.trace import Accuracy, ModelCheckpoint
+from fastestimator.estimator.trace import Accuracy, LRController, ModelCheckpoint
 from fastestimator.network.loss import SparseCategoricalCrossentropy
+from fastestimator.network.lrschedule import CyclicLRSchedule
 from fastestimator.network.model import FEModel, ModelOp
 from fastestimator.pipeline.processing import Minmax
 
@@ -42,7 +43,7 @@ def get_estimator(epochs=2, batch_size=32):
     # step 3.prepare estimator
     traces = [
         Accuracy(true_key="y", pred_key="y_pred", output_name='acc'),
-        ModelCheckpoint('./lenet_model', monitor_name='acc', save_best_only=True)
+        LRController(model_name="lenet", lr_schedule=CyclicLRSchedule())
     ]
     estimator = fe.Estimator(network=network, pipeline=pipeline, epochs=epochs, traces=traces)
     return estimator
--- a/fastestimator/estimator/estimator.py
+++ b/fastestimator/estimator/estimator.py
@@ -161,7 +161,7 @@ class Estimator:
 
     def _start(self):
         self.train_step = 0
-        self._run_traces_on_begin({"num_devices": self.num_devices})
+        self._run_traces_on_begin({"train_step": self.train_step, "num_devices": self.num_devices})
         for self.train_epoch in range(self.epochs):
             self._run_epoch("train")
             if self.do_eval:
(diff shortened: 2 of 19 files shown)


fastestimator/estimator/estimator.py
```diff
@@ -143,13 +148,24 @@ def _sort_traces(self):
...
             state = {"mode": mode}
-            for epoch in signature_epochs:
+            for idx, epoch in enumerate(signature_epochs):
                 ds_iter = self.pipeline.dataset_schedule[mode].get_current_value(epoch)
+                if mode == "train":
+                    global_batch_size = self.pipeline.get_global_batch_size(epoch)
+                    if self.steps_per_epoch:
+                        max_steps = self.steps_per_epoch
+                    else:
+                        max_steps = min(self.pipeline.num_examples[mode]) // global_batch_size
```
TortoiseHam: do we do padding on the last batch in an epoch to fill it out? if so, don't we need to round up here?

vbvg2008: no padding

fastestimator/estimator/trace.py
```diff
@@ -195,6 +195,95 @@ def on_epoch_end(self, state):
...
+
+
+class LRController(Trace):
+    def __init__(self,
+                 model,
+                 lr_schedule=None,
+                 reduce_on_eval=False,
+                 reduce_patience=10,
+                 reduce_factor=0.1,
+                 reduce_metric="loss",
```
TortoiseHam: why separate reduce_on_eval from reduce_metric? we can just infer reduce_on_eval from whether a non-None reduce_metric is provided

fastestimator/estimator/trace.py
```diff
@@ -195,6 +195,95 @@ def on_epoch_end(self, state):
...
+        self.change_lr = False
+        self.wait = 0
+        if self.lr_schedule:
+            assert isinstance(self.lr_schedule, LRSchedule), "lr_schedule must be instance of LRSchedule"
+        if self.reduce_metric_mode == "min":
+            self.reduce_metric_best = np.Inf
+        elif self.reduce_metric_mode == "max":
+            self.reduce_metric_best = -np.Inf
+        else:
+            raise ValueError("reduce_metric_mode is either 'min' or 'max'")
```
TortoiseHam: "must be either 'min' or 'max' "

fastestimator/estimator/trace.py
```diff
@@ -195,6 +195,95 @@ def on_epoch_end(self, state):
...
+            current_value = state[self.reduce_metric]
+            if self._is_better(current_value):
+                self.reduce_metric_best = current_value
+                self.wait = 0
+            else:
+                self.wait += 1
+                if self.wait >= self.reduce_patience:
+                    self.reduce_lr_ratio *= self.reduce_factor
+                    self.change_lr = True
+                    print("FastEstimator-LearningRateChanger: learning rate reduced by factor of {}".format(
```
TortoiseHam: "FastEstimator-LRController..."

fastestimator/estimator/trace.py
```diff
@@ -792,7 +821,7 @@ def on_epoch_end(self, state):
                     for name, model in self.network.model.items():
                         model.set_weights(self.best_weights[name])
                 print("FastEstimator-EarlyStopping: '{}' triggered an early stop. Its best value was {} at epoch {}\
-                        ".format(self.monitored_key, self.best, state['epoch'] - self.wait))
+                        "                         .format(self.monitored_key, self.best, state['epoch'] - self.wait))
```
TortoiseHam: what's going on with this?

fastestimator/network/lrschedule.py
```diff
@@ -0,0 +1,56 @@
+import math
+
+import numpy as np
+
+
+class LRSchedule:
+    def __init__(self, schedule_mode):
+        self.schedule_mode = schedule_mode
+        self.total_epochs = None  #will be filled on runtime
```
TortoiseHam: "will be filled by runtime"

vbvg2008: fixed

fastestimator/estimator/trace.py
vbvg2008: solar flare

fastestimator/estimator/trace.py
vbvg2008: fixed

fastestimator/estimator/trace.py
vbvg2008: fixed

fastestimator/estimator/trace.py
vbvg2008: fixed.

apphub/image_classification/lenet_mnist.py
```diff
@@ -42,7 +43,7 @@ def get_estimator(epochs=2, batch_size=32):
     # step 3.prepare estimator
     traces = [
         Accuracy(true_key="y", pred_key="y_pred", output_name='acc'),
-        ModelCheckpoint('./lenet_model', monitor_name='acc', save_best_only=True)
+        LRController(model_name="lenet", lr_schedule=CyclicLRSchedule())
```
TortoiseHam: if you're removing ModelCheckpoint, need to clean up the import statement

vbvg2008: not removing, i'm currently cleaning up the checkpoint code

fastestimator/estimator/trace.py
```diff
@@ -195,6 +195,97 @@ def on_epoch_end(self, state):
...
+class LRController(Trace):
+    def __init__(self,
+                 model_name,
+                 lr_schedule=None,
+                 reduce_on_eval=False,
+                 reduce_patience=10,
+                 reduce_factor=0.1,
+                 reduce_mode="min",
+                 min_lr=1e-6):
+        if reduce_on_eval and isinstance(reduce_on_eval, str):
```
TortoiseHam: this can just be 
`if is instance(reduce_on_eval, str)`

TortoiseHam: or even get rid of the outer if and do

super().__init__(inputs=reduce_on_eval if isinstance(reduce_on_eval,str) else None)

vbvg2008: fixed

fastestimator/estimator/trace.py
```diff
@@ -1045,27 +1076,25 @@ class ModelCheckpoint(Trace):
...
         mode (str): Save models during either `train` or `eval`. Default is `eval`.
-        verbose (int): verbosity mode, 0 or 1
         save_best_only (bool): Keep only the best model.
         save_mode: Can be `'min'`, `'max'`, or `'auto'`.
         save_freq: Number of epochs to save models. Cannot be used with `save_best_only=True`.
     """
     def __init__(self,
                  save_dir,
-                 model_names=None,
+                 models,
```
TortoiseHam: is models really more intuitive than model_names here? if so, the doc string needs to say 'models' rather than 'model'

fastestimator/estimator/trace.py
```diff
@@ -1045,27 +1076,25 @@ class ModelCheckpoint(Trace):
...
                  monitor_name=None,
                  mode='eval',
                  save_best_only=False,
                  save_mode='auto',
-                 save_freq=1,
-                 verbose=1):
```
TortoiseHam: verbose was probably here to match the keras method signature. how much to we care about keeping that the same? if we don't care we can probably simplify some other traces too

fastestimator/network/lrschedule.py
```diff
@@ -0,0 +1,56 @@
...
+        self.total_epochs = None  #will be filled by runtime
+        self.total_steps = None  #will be filled by runtime
+        self.initial_lr = None  #will be filled by runtime
+
+    def schedule_fn(self, current_step_or_epoch, lr):
+        #do something to change lr
+        return lr
+
+
+class CyclicLRSchedule(LRSchedule):
```
TortoiseHam: do you think it would be worth implementing the math functions here as @tf.function since they might get invoked at every step?

fastestimator/estimator/trace.py
```diff
@@ -195,6 +195,97 @@ def on_epoch_end(self, state):
...
+                    self.change_lr = True
+                    print("FastEstimator-LRController: learning rate reduced by factor of {}".format(
+                        self.reduce_factor))
+
+    def _update_lr(self):
+        self.current_lr = max(self.base_lr * self.reduce_lr_ratio, self.min_lr)
+        backend.set_value(self.model.optimizer.lr, self.current_lr)
+        self.change_lr = False
+
+    def _is_better(self, value):
```
TortoiseHam: this could be done by selecting the comparison op up front when you're setting reduce_metric_best in the init method. see the early stopping trace for example

fastestimator/estimator/trace.py
```diff
@@ -195,6 +195,97 @@ def on_epoch_end(self, state):
...
+    def on_batch_begin(self, state):
+        if state["mode"] == "train":
+            if self.lr_schedule and self.lr_schedule.schedule_mode == "step":
+                self.base_lr = self.lr_schedule.schedule_fn(state["train_step"], self.base_lr)
+                self.change_lr = True
+            if self.change_lr:
+                self._update_lr()
+
+    def on_batch_end(self, state):
+        if state["mode"] == "train" and self.log_steps and state["train_step"] % self.log_steps == 0:
```
TortoiseHam: is there any reason to only emit this on log_steps? doing this if check might be just as expensive as just always putting it in the dictionary

apphub/image_classification/lenet_mnist.py
TortoiseHam: but you took it out of the trace list here. either it should be put back or the unused import should be removed

fastestimator/estimator/trace.py
vbvg2008: yes, the performance improved when only emit this on log_steps

also, we don't want unnecessary stuff to be included in the state dictionary if they are not going to be be used anywhere.

fastestimator/estimator/estimator.py
```diff
@@ -80,6 +82,11 @@ def _prepare_estimator(self):
             assert isinstance(trace, Trace)
             if isinstance(trace, ModelCheckpoint):
                 no_save_warning = False
+            if isinstance(trace, LRController):
+                trace.log_steps = self.log_steps
+                if trace.lr_schedule:
+                    trace.lr_schedule.total_epochs = self.epochs
```
TortoiseHam: instead of manually setting this from the estimator, we should pass these variables in the on_begin trace dictionary. that way other users who have traces that want this info can get it too

TortoiseHam: likewise with the log_steps info for all traces. better to minimize the amount of coupling between estimator and specific traces

fastestimator/estimator/estimator.py
TortoiseHam: are we throwing out the last partial batch?

fastestimator/estimator/trace.py
TortoiseHam: huh. interesting

fastestimator/estimator/estimator.py
vbvg2008: good idea, fixed

fastestimator/estimator/estimator.py
vbvg2008: yes.

apphub/image_classification/lenet_mnist.py
vbvg2008: fixed

fastestimator/estimator/trace.py
vbvg2008: fixed.

fastestimator/estimator/trace.py
vbvg2008: yeah, we don't need to do the same, because we are already different from keras.

fastestimator/network/lrschedule.py
vbvg2008: i don't think so, the computation is just a scalar, cpu is better for it.   

gpu is good for large data when data parallelism can be used.

fastestimator/estimator/trace.py
vbvg2008: done

fastestimator/estimator/trace.py
vbvg2008: yup

PR #170, opened 2019-09-18
added docstring for pipeline

purujitb:


fastestimator/pipeline/pipeline.py
```diff
@@ -30,6 +30,18 @@
 
 
 class Pipeline:
+    """
+        Class representing the data pipeline required for fastestimator
```
chm123: There is no line break between The """ and the summary.

```python
"""Class representing the data pipeline required for fastestimator.
```

PR #181, opened 2019-09-20
added docstring for network

Vivek305:


fastestimator/network/lrschedule.py
```diff
@@ -4,6 +4,9 @@
 
 
 class LRSchedule:
+    """A base class for learning rate scheduler.
+
```
vbvg2008: add schedule mode explanation

fastestimator/network/lrschedule.py
```diff
@@ -16,6 +19,13 @@ def schedule_fn(self, current_step_or_epoch, lr):
 
 
 class CyclicLRSchedule(LRSchedule):
+    """A class representing cyclic learning rate scheduler
+
+    Args:
+        num_cycle: The number of cycles to be used by the learning rate scheduler
+        cycle_multiplier: The length of each next cycle's multiplier
```
vbvg2008: multiplier of next cycle length w.r.t previous cycle length

PR #186, opened 2019-09-27
lung segmentation tutorial

rajesh1226:
This is the tutorial demonstrating lung segmentation training using UNet over montgomery dataset.


chm123:
Remove the code block if it is commented out.

PR #208, opened 2019-10-02
added-configs-for-unet

hanskrupakar:


fastestimator/architecture/unet.py
```diff
@@ -12,66 +12,151 @@
...
+
+    # Expanding blocks
+    inp1, inp2 = levels[-1][1], levels[-1][0]
+    for idx, nc in enumerate(reversed(nchannels[1:])):
+        if idx == 0:
+            d = dropout
+        else:
+            d = None
+        
+        D = up_concat(inp1, inp2, 2, (nchannels[-1-idx], nchannels[-2-idx]), 3, conv_config, d, bn=bn)
```
TortoiseHam: does dropout go in both the up and down? I only see it in one place in the old file

PR #210, opened 2019-10-03
deleting FEModel, now it's time for fe.build

vbvg2008:
`fe.build` will take one definition and can return multiple models

for single model definition, simply replace FEModel with fe.build


fastestimator/network.py
```diff
@@ -181,3 +174,62 @@ def _reduce_loss(self, batch, global_batch_size, epoch_losses, warm_up):
...
+    """
+    convert_tuple = False
+    if get_num_devices() > 1:
+        distribute_strategy = tf.distribute.MirroredStrategy()
+    else:
+        distribute_strategy = None
+    with distribute_strategy.scope() if distribute_strategy else NonContext():
+        model = model_def()
+        if isinstance(model, (list, tuple)):
+            assert len(model) == len(model_name) == len(optimizer) == len(loss_name)
```
TortoiseHam: we need to call to_list on model_name, optimizer, loss_name first. else could end up fooled with 4 models and loss_name="loss"

fastestimator/network.py
```diff
@@ -181,3 +174,62 @@ def _reduce_loss(self, batch, global_batch_size, epoch_losses, warm_up):
...
+    Args:
+        model_def (function): function definition of tf.keras model
+        model_name (str, list, tuple): model name(s)
+        optimizer (str, optimizer, list, tuple): optimizer(s)
+        loss_name (str, list, tuple): loss name(s)
+
+    Returns:
+        model: model(s) compiled by FastEstimator
+    """
+    convert_tuple = False
```
TortoiseHam: instead of going to all the hassle to maintain the tuple vs list distinction, lets just call to_list on everything right up front

TortoiseHam: this also lets us get rid of the if-branching at line 198

vbvg2008: what we need is to return the same data structure that user defines in the model_def

fastestimator/network.py
```diff
@@ -181,3 +174,62 @@ def _reduce_loss(self, batch, global_batch_size, epoch_losses, warm_up):
                     "please make sure loss is element-wise loss"
             reduced_loss[loss_name] = tf.reduce_sum(element_wise_loss) / global_batch_size
         return reduced_loss
+
+
+def build(model_def, model_name, optimizer, loss_name):
```
TortoiseHam: if we're getting rid of loss_name="loss" (which I'm still not sold on) then shouldn't we also change all of the loss files to not have that default output name?

vbvg2008: if we are getting rid of default loss_name, then yes!

fastestimator/network.py
vbvg2008: fixed

fastestimator/network.py
vbvg2008: fixed

PR #222, opened 2019-10-05
Refactor dataset API

chm123:
- Clean up docstring
- Chang download path to `fastestimator_data` under user's home directory
- Display progress bar in terminal and jupyter notebook

### DIFF
--- a/fastestimator/dataset/horse2zebra.py
+++ b/fastestimator/dataset/horse2zebra.py
@@ -27,13 +27,13 @@ from fastestimator.util.wget import bar_custom, callback_progress
 wget.callback_progress = callback_progress
 
 
-def _create_csv(img, key_name, parent_path, csv_path):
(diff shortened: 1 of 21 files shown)


fastestimator/dataset/horse2zebra.py
```diff
@@ -12,66 +12,87 @@
...
+            path (str): Path to data root directory.
+
     """
+    home = str(Path.home())
+
     if path is None:
-        path = os.path.join(tempfile.gettempdir(), ".fe", "HORSE2ZEBRA")
-    if not (os.path.exists(path)):
-        os.makedirs(path)
+        path = os.path.join(home, 'fastestimator_data', 'horse2zebra')
```
TortoiseHam: should we extract all of this path management into a common utility function so that no one does it wrong in the future?

vbvg2008: you r right!   i will make the change over weekend

PR #235, opened 2019-10-09
added-residual-connections-to-unet

hanskrupakar:


fastestimator/architecture/unet_3d.py
```diff
@@ -64,9 +66,13 @@ def UNet3D(input_size=(9, 512, 512, 1),
 
     assert len(nchannels) == len(dilation_rates), "len(nchannels) should be the same as len(dilation_rates)"
 
+    assert isinstance(residual, bool), 'Residual argument can be boolean only'
```
TortoiseHam: why boolean only here when the other allowed 'enc' and 'dec' too?

hanskrupakar: That's a design choice. The 3D UNet model doesn't use `padding='same'` by default and so the sizes are only consistent across the concatenation layers. To make the residual connections easier to manage, I use `padding='same'` if `residual=True` and this makes it impossible to use the same with one of encoder or decoder layers as it will mess with the concatenation operation.

PR #237, opened 2019-10-09
separate out the test script and added tutorial tests

Vivek305:


test/test_tutorials.sh
```diff
@@ -0,0 +1,23 @@
...
+    jupyter nbconvert --to script ${filename} --output ${path_temp}${fname}
+    if ipython ${path_temp}${fname}'.py'; then
+        ((cnt=cnt+1))
+    else
+        echo 'failed '${fname}
+        exit 0
+    fi
+done
+rm ${path_temp}/*.py
+echo $cnt 'tests passed out of 10 tests'
```
TortoiseHam: can we infer the total number dynamically instead of hard-coding 10?

Vivek305: Yes. updated the script

test/test_image_classification/test_densenet121.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+path_apphub='../apphub/'
```
vbvg2008: instead of using "../apphub/",  let's automatically infer the path like this 
 https://github.com/fastestimator/fastestimator/blob/r0.2/test/tests.sh

then we don't need to go to the test folder to execute the bash

PR #242, opened 2019-10-10
added NIH Chest Xray dataset, 112120 images

vbvg2008:


fastestimator/dataset/nih_chestxray.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+import tarfile
+from pathlib import Path
+
+import pandas as pd
+import wget
+
+from fastestimator.util.wget import bar_custom, callback_progress
+
+wget.callback_progress = callback_progress
```
TortoiseHam: why is this outside of the function scope?

PR #246, opened 2019-10-11
RecordWriter Speed

TortoiseHam:
first pass at speeding up recordwriter. Don't merge yet, but looking for speed comparison on different systems


TortoiseHam:
Ok, this is ready for merging now unless anyone has objections

PR #255, opened 2019-10-15
PGGAN refactored

jphong89:


apphub/image_generation/pggan_nihchestxray/pggan_nihchestxray.py
```diff
@@ -152,22 +151,28 @@ def on_epoch_end(self, state):
...
-            plt.savefig(
-                os.path.join(self.save_dir,
-                             'image_at_{:08d}.png').format(state["epoch"]))
-            print("on epoch {}, saving image to {}".format(
-                state["epoch"], self.save_dir))
-
-
-def get_estimator():
-    train_csv, data_path = load_data()
+                if self.num_channels == 1:
```
vbvg2008: what happens when the `self.num_channels` is not 1?   I think there may be some indentation issue.

jphong89: Oh, you are right. Indentation seems to be messed up.. lemme fix this

PR #289, opened 2019-10-21
Add slack notification trace

chm123:


fastestimator/trace/io/slackio.py
```diff
@@ -0,0 +1,75 @@
...
+        if self.begin_msg:
+            self._send_message(self.begin_msg)
+
+    def on_end(self, state):
+        if isinstance(self.end_msg, str):
+            self._send_message(self.end_msg)
+        elif isinstance(self.end_msg, types.FunctionType):
+            self._send_message(self.end_msg(state))
+        else:
+            raise TypeError("end_msg must be string or function.")
```
TortoiseHam: probably would want to raise this at the start rather than the end

fastestimator/trace/io/slackio.py
```diff
@@ -0,0 +1,75 @@
...
+        try:
+            self.client.chat_postMessage(channel=self.channel, text=msg)
+        except SlackApiError as err:
+            if self.verbose > 0:
+                print(err)
+            print("Slack API error. Continue training without sending Slack notification.")
+
+    def on_begin(self, state):
+        if self.begin_msg:
+            self._send_message(self.begin_msg)
```
TortoiseHam: should we allow a function here too?

fastestimator/trace/io/slackio.py
chm123: Moved to __init__

fastestimator/trace/io/slackio.py
chm123: Done.

PR #343, opened 2019-11-13
Add PredictBox class for retinanet

chm123:


chm123:
> tf.image.non_max_suppression has a parameter **score_threshold**. If we use that it will see that among the input boxes , boxes having score of more than 0.05 shall we selected.
> It would be concise as well.

The default nms iou threshold is 0.5, the same as RetinaNet paper. 0.05 is not for nms.

PR #362, opened 2019-11-21
Adding siamese_network showcase

gaurav272333:


apphub/one_shot_learning/siamese_network.py
```diff
@@ -0,0 +1,244 @@
...
+import numpy as np
+import tensorflow as tf
+from tensorflow.python.keras import Model, layers, Sequential
+from tensorflow.python.keras.initializers import RandomNormal
+from tensorflow.python.keras.regularizers import l2
+from tensorflow.keras.optimizers import Adam
+
+import fastestimator as fe
+from fastestimator.op import TensorOp
+from fastestimator.trace import Trace, LRController, Accuracy, ModelSaver, EarlyStopping
```
TortoiseHam: unused import

apphub/one_shot_learning/siamese_network.py
```diff
@@ -0,0 +1,244 @@
...
+from fastestimator.schedule import LRSchedule
+from fastestimator.op.tensorop import ModelOp, BinaryCrossentropy, Minmax, Augmentation2D
+
+from fastestimator.dataset.omniglot import load_data, load_eval_data, get_batch, one_shot_trial
+
+
+class ConditionalAugmentation(TensorOp):
```
TortoiseHam: we should make a more general Op wrapper that can make any op conditional

TortoiseHam: possibly part of the base class?

TortoiseHam:
Does it make sense for "one_shot_learning" to be it's own app hub folder? or should this be inside of image_classification or whatever it's actually doing?

apphub/one_shot_learning/siamese_network.py
gaurav272333: Sure. I will remove that.

apphub/one_shot_learning/siamese_network.py
vbvg2008: I implemented a general wrapper for that

PR #400, opened 2020-01-09
Bring Pipeline to 1.0dev

TortoiseHam:
fix tensorflow apphub example (+2 squashed commits)
Squashed commits:
[63f58a6] fix estimator to work with pipeline
[03a6664] fix a bunch of formatting problems, bring future pipeline into 1.0dev


setup.py
```diff
@@ -46,10 +46,11 @@ def get_name():
     keywords="fastestimator tensorflow",
     classifiers=[
         "License :: OSI Approved :: Apache Software License",
-        "Programming Language :: Python :: 3", ],
+        "Programming Language :: Python :: 3",
+    ],
 
     # Declare minimal set for installation
-    install_requires=['numpy'],
+    install_requires=['numpy', 'torch', 'tensorflow', 'albumentations'],
```
vbvg2008: torch should not be here, because `pip install torch` will not work for windows users.

for consistency, let's not put tensorflow here either.

apphub/image_classification/lenet_mnist/pytorch/lenet_mnist.py
```diff
@@ -6,35 +20,14 @@
...
-from torch.utils.data import DataLoader, Dataset
 
 import fastestimator as fe
+from fastestimator.pipeline import Pipeline, NumpyDataset
+from fastestimator.op.numpyop import Minmax, ExpandDims
 from fastestimator.op.tensorop.loss import CrossEntropy
 from fastestimator.op.tensorop.model import ModelOp, UpdateOp
 from fastestimator.trace.metric import Accuracy
 
-class MnistDataset(Dataset):
```
vbvg2008: can you create a new file in the pytorch folder instead of over-writing the old one?   we need a file that showcases torch data loader in pipeline for now.

apphub/image_classification/lenet_mnist/tensorflow/lenet_mnist.py
```diff
@@ -42,28 +58,38 @@ def LeNet(input_shape=(28, 28, 1), classes=10):
     return model
 
 
-def get_estimator():
```
vbvg2008: same as before, let's create a separate file in tensorflow folder.

setup.py
TortoiseHam: but pip install tensorflow will finally work now :(

TortoiseHam: the torch website seems to think pip will work on windows

apphub/image_classification/lenet_mnist/pytorch/lenet_mnist.py
TortoiseHam: realistically we're not going to have 4 different examples just to show off these permutations... I think in app hub we probably don't want any examples that aren't using our pipeline since we want users to do it our way. That would be something better covered in a tutorial

vbvg2008: yes, we will move it to tutorial later in the future when we re-write the tutorials. For now, let's have two files for this example only.

TortoiseHam: added the extra files for now

apphub/image_classification/lenet_mnist/tensorflow/lenet_mnist.py
TortoiseHam: added the extra files for now

PR #407, opened 2020-01-22
refactored pipeline

vbvg2008:
temporarily getting rid of scheduling related codes, since we are going to rewrite the schedule module


fastestimator/__init__.py
```diff
@@ -14,6 +14,6 @@
 # ==============================================================================
 __version__ = '1.0'
 
-from fastestimator.pipeline import Pipeline
-from fastestimator.network import Network, build
 from fastestimator.estimator import Estimator
+from fastestimator.network import Network, build
+from fastestimator.pipeline import Pipeline
```
TortoiseHam: i think Pipeline import might need to precede Estimator since Estimator imports Pipeline for type checking

fastestimator/estimator.py
```diff
@@ -46,43 +45,21 @@ class Estimator:
         log_steps (int, optional): Interval steps of logging. Defaults to 100.
         monitor_names (str, list): Additional keys to print in logger
     """
-    pipeline: Scheduler[BasePipeline]
+    pipeline: Union[FEPipeline, TensorFlowPipeline, TorchPipeline]
```
TortoiseHam: this should just be BasePipeline instead of Union

fastestimator/estimator.py
```diff
@@ -46,43 +45,21 @@ class Estimator:
...
     epochs: int
     steps_per_epoch: Optional[int]
     traces: List[Trace]
     log_steps: int
 
     def __init__(self,
-                 pipeline: Union[BasePipeline, Dict[str, Union[DataLoader, tf.data.Dataset]], Scheduler[BasePipeline]],
+                 pipeline: Union[FEPipeline, TensorFlowPipeline, TorchPipeline],
```
TortoiseHam: this should just be BasePipeline instead of Union

fastestimator/estimator.py
```diff
@@ -102,18 +79,15 @@ def fit(self):
...
-                mode="train", epoch=0)
+            self.steps_per_epoch = self.pipeline.get_num_steps(mode="train", epoch=0)
         self.system.total_steps = self.epochs * self.steps_per_epoch
 
     def _prepare_network(self):
         self.network.exported_keys = self.network.op_outputs.intersection(self.trace_inputs)
 
     def _prepare_estimator(self):
-        self.do_eval = "eval" in self.pipeline.get_current_value(0).get_modes()  # TODO - Scheduling
+        self.do_eval = self.pipeline.eval_data is not None
```
TortoiseHam: probably shouldn't reference internal variables like this directly. abstract class paradigm should define a relevant method instead

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
-        raise NotImplementedError()
-
-    def get_global_batch_size(self, mode: str, epoch: int) -> int:
-        raise NotImplementedError()
-
-    def get_signature_epochs(self, mode: str) -> List[int]:
-        raise NotImplementedError()
-
-    def get_all_output_keys(self) -> Set[str]:
+    def __init__(self,
```
TortoiseHam: basePipeline isn't the right place to specify the init since then anyone extending it for a new pipeline with different data types would have to edit the BasePipeline too

fastestimator/pipeline.py
```diff
@@ -15,14 +15,14 @@
 import os
 import sys
```
TortoiseHam: unused imports

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
+                 eval_data: Union[Dataset, DataLoader, Scheduler, tf.data.Dataset, None]):
+        self.train_data = train_data
+        self.eval_data = eval_data
+
+    def get_batch_size(self, mode: str, epoch: int) -> int:
+        return None
+
+    def get_iterator(self, mode: str, epoch: int) -> Iterator:
+        itr = None
+        if mode == "train":
```
TortoiseHam: this implementation is inconsistent with the BasePipeline init args - which are not all directly iterators

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
-    def get_modes(self) -> Set[str]:
-        raise NotImplementedError()
-
-    def get_global_batch_size(self, mode: str, epoch: int) -> int:
-        raise NotImplementedError()
-
-    def get_signature_epochs(self, mode: str) -> List[int]:
-        raise NotImplementedError()
-
-    def get_all_output_keys(self) -> Set[str]:
```
TortoiseHam: what is your plan for later validating whether traces will plausibly work given their input requirements?

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
+    def __init__(self,
+                 train_data: Union[Dataset, DataLoader, Scheduler, tf.data.Dataset],
+                 eval_data: Union[Dataset, DataLoader, Scheduler, tf.data.Dataset, None]):
+        self.train_data = train_data
+        self.eval_data = eval_data
+
+    def get_batch_size(self, mode: str, epoch: int) -> int:
+        return None
```
TortoiseHam: why None rather than NotImplementedError ?

TortoiseHam: also this conflicts with the method signature which promises an int return type

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
                 start = time.perf_counter()
 
-    def show_results(self, mode: str = "train", num_steps: int = 1, epoch: int = 0):
-        data = []
-        itr = self.transform(mode=mode, epoch=epoch)
-        for _ in range(num_steps):
-            data.append(next(itr))
-        return data
 
+class TorchPipeline(BasePipeline):
```
TortoiseHam: failed to implement all BasePipeline methods

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
+
+    def get_num_steps(self, mode: str, epoch: int) -> int:
         raise NotImplementedError()
 
-    def get_num_examples(self, mode: str, epoch: int) -> int:
+    def index(self, idx: int, mode: str = "train", epoch: int = 0) -> dict:
         raise NotImplementedError()
 
-    def transform(self, mode: str, epoch: int = 0, dataset: Optional[Dataset] = None) -> Iterator:
+    def transform(self, data: dict) -> dict:
```
TortoiseHam: what is the point of the transform method now? It doesn't seem to be implemented anywhere

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
+            itr = self.train_data
+        elif mode == "eval":
+            itr = self.eval_data
+        return itr
+
+    def get_num_steps(self, mode: str, epoch: int) -> int:
         raise NotImplementedError()
 
-    def get_num_examples(self, mode: str, epoch: int) -> int:
+    def index(self, idx: int, mode: str = "train", epoch: int = 0) -> dict:
```
TortoiseHam: the idea is to use this instead of show_results right?

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
+        dataloader = self.get_iterator(mode=mode, epoch=epoch)
+        return len(dataloader)
+
+    def index(self, idx: int, mode: str = "train", epoch: int = 0) -> dict:
+        dataloader = self.get_iterator(mode=mode, epoch=epoch)
+        dataset = dataloader.dataset
+        return dataset[idx]
+
+
+class TensorFlowPipeline(BasePipeline):
```
TortoiseHam: need to implement all base class functions

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
-    # TODO support filter ops
-    # TODO support cache op
-    dataloaders: Dict[str, Scheduler[DataLoader]]
-    batch_size: Scheduler[int]
-    """ A class representing the data pipeline for FastEstimator
+    def get_num_steps(self, mode: str, epoch: int) -> int:
+        dataset = self.get_iterator(mode=mode, epoch=epoch)
+        num_steps = tf.data.experimental.cardinality(dataset)
+        if num_steps in [tf.data.experimental.INFINITE_CARDINALITY, tf.data.experimental.UNKNOWN_CARDINALITY]:
+            raise ValueError("Cannot infer total steps, please provide 'steps_per_epoch' in Estimator")
```
TortoiseHam: what was wrong with the other mechanism of inferring these values?

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
-                 train_data: Optional[Dataset] = None,
-                 eval_data: Optional[Dataset] = None,
-                 batch_size: Union[int, Scheduler] = 1,
+                 train_data: Union[Dataset, Scheduler],
+                 eval_data: Union[Dataset, Scheduler, None] = None,
+                 batch_size: Union[int, Scheduler, None] = None,
                  ops: Union[None, NumpyOp, Scheduler[NumpyOp], Iterable[Union[NumpyOp, Scheduler[NumpyOp]]]] = None,
-                 drop_last: bool = False,
-                 num_process: Optional[Union[int, Scheduler[int]]] = None,
-                 shuffle_train: bool = True,
```
TortoiseHam: no reason to remove user flexibility to set these values

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
+        self.num_process = os.cpu_count() if num_process is None else num_process
         self.datasets = {}
         self.dataloaders = {}
-        if train_data is not None:
-            self.datasets["train"] = self._build_dataset(train_data, "train")
-            # Intentionally not using self.num_process in the build_loader call since want to trigger perf test
-            self.dataloaders["train"] = self._build_loader(self.datasets["train"],
-                                                           shuffle=shuffle_train,
-                                                           num_process=num_process)
+        self.datasets["train"] = self._build_dataset(train_data, "train")
```
TortoiseHam: train_data should be optional so that as a user with a trained model i can still instantiate a pipeline only for testing or validation data

fastestimator/pipeline.py
```diff
@@ -243,71 +252,27 @@ def _build_loader(self,
...
-            elif best == 2:
-                epoch_dict[epoch] = OpDataLoader(dataset,
-                                                 batch_size=size,
-                                                 shuffle=shuffle,
-                                                 num_workers=os.cpu_count(),
-                                                 drop_last=self.drop_last)
-        return Scheduler(epoch_dict=epoch_dict)
+            best_idx = times.index(min(times))
+            num_process = cpu_list[best_idx]
+        elif num_process < 6:
```
TortoiseHam: if <6 set to zero??

fastestimator/pipeline.py
```diff
@@ -243,71 +252,27 @@ def _build_loader(self,
...
-                                                 drop_last=self.drop_last)
-            elif best == 2:
-                epoch_dict[epoch] = OpDataLoader(dataset,
-                                                 batch_size=size,
-                                                 shuffle=shuffle,
-                                                 num_workers=os.cpu_count(),
-                                                 drop_last=self.drop_last)
-        return Scheduler(epoch_dict=epoch_dict)
+            best_idx = times.index(min(times))
+            num_process = cpu_list[best_idx]
```
TortoiseHam: this is good - much cleaner than the if-else blocks I was using

fastestimator/pipeline.py
TortoiseHam: unless i'm missing something i don't think TF datasets support fetch by index

fastestimator/pipeline.py
```diff
@@ -139,102 +139,111 @@ def get_per_epoch(ops: Iterable[Union[T, Scheduler[T]]], epoch: int) -> List[T]:
...
+
+    def get_iterator(self, mode: str, epoch: int) -> Iterator:
+        itr = None
+        if mode == "train":
+            itr = self.train_data
+        elif mode == "eval":
+            itr = self.eval_data
+        return itr
+
+    def get_num_steps(self, mode: str, epoch: int) -> int:
```
TortoiseHam: I like having the get_num_steps call here. I wonder if we should keep get_num_examples too though for users to have easier time debugging / inspecting their pipelines

fastestimator/pipeline.py
```diff
@@ -317,146 +282,35 @@ def _test_loader(self, dataset: OpDataset, batch_size: int, shuffle: bool, num_w
...
+        return dataset[idx]
+
+
+def Pipeline(train_data: Union[Dataset, DataLoader, Scheduler, tf.data.Dataset],
+             eval_data: Union[Dataset, DataLoader, Scheduler, tf.data.Dataset, None] = None,
+             batch_size: Union[int, Scheduler, None] = None,
+             ops: Union[None, NumpyOp, Scheduler[NumpyOp], Iterable[Union[NumpyOp, Scheduler[NumpyOp]]]] = None,
+             num_process: Union[int, None] = None):
+    assert isinstance(train_data, (Dataset, DataLoader, Scheduler, tf.data.Dataset))
+    if not eval_data:
```
TortoiseHam: why is the type check only if there is not eval_data?

fastestimator/pipeline.py
```diff
@@ -317,146 +282,35 @@ def _test_loader(self, dataset: OpDataset, batch_size: int, shuffle: bool, num_w
...
+        assert type(train_data) == type(eval_data)
```
TortoiseHam: this check needlessly prevents someone from, for example, having NumpyData for training but GeneratorData for testing

PR #410, opened 2020-02-04
add new apphub examples

geez0219:
1. add MAML, siamese_network and uncertainty_loss_cub200 totally 3 examples into nightly-build testing routine

2. enhance the nightly-build test running such that it will show the error message directly when any example or tutorial fails. 

3. correct the cyclegan path in apphub readme.


test/run_nightly_build.py
```diff
@@ -33,14 +38,19 @@
         for f in filenames:
             if f.split(".")[-1] != "py":
                 continue
-
+
```
chm123: Remove trailing spaces

PR #425, opened 2020-02-07
Fix multigpu apphub bug

chm123:
This turns off distribution strategy in the inferencing section to avoid error.


geez0219:
The pr-merge test fail is due to Jenkins pipeline's problem. 
This problem should be solved after #427 being merged.

PR #434, opened 2020-02-10
FastEstimator New API Introduction

vbvg2008:
This is not the finished version of our new API,  it is a starting point that can achieve all of the following improvements. I recommend merge this PR first then we can all discuss it.

Key improvements (comparing new API to 1.0b)

* Overall:
1. Both TF 2.1.0 and Pytorch 1.3.1 users are welcome
2. now notebook friendly
3. less API, more intuitive

* Pipeline
2. Pipeline is now pure numpy! good for debugging!
3. pipeline is purely dynamic (no disk writing/reading anymore)
4. pipleine can still take tf.dataset and torch.loader
5. can use pipeline for a lot more than training itself (data generation, model testing ...)

* Network:
1. better data memory control:  CPU -> control -> GPU ->control -> CPU

* Estimator:
1. Trace now has inputs,
2. Key check (including checking Trace)
3. Trace now has a unified system object to
4. we still have warm up

* Schedule:
3 Different kinds of scheduling available


fastestimator/dataset/fe_dataset.py
```diff
@@ -0,0 +1,38 @@
...
+# limitations under the License.
+# ==============================================================================
+
+from torch.utils.data import Dataset
+
+from fastestimator.op import get_inputs_by_op, write_outputs_by_key
+
+
+class OpDataset(Dataset):
+    def __init__(self, dataset, ops, mode):
```
TortoiseHam: needs typing

fastestimator/estimator.py
```diff
@@ -35,27 +36,26 @@ class Estimator:
...
         epochs (int): Number of epooch to run.
-        steps_per_epoch (int, optional): Number of steps to run for each epoch.
+        steps_per_epoch (int, optional): maximum steps to run for each epoch. If None, all data will be used
         traces (list, optional): List of the traces objects to run during training. If None, there will be only basic
             traces.
         log_steps (int, optional): Interval steps of logging. Defaults to 100.
         monitor_names (str, list): Additional keys to print in logger
     """
-    pipeline: BasePipeline
+    pipeline: Pipeline
```
TortoiseHam: should either be 'Pipeline' and 'Network' or 'BasePipeline' and 'BaseNetwork'

fastestimator/estimator.py
```diff
@@ -64,92 +64,104 @@ def __init__(self,
...
-        for op in self.network.ops:
-            if isinstance(op, UpdateOp):
-                loss_keys = loss_keys.union(set(to_list(op.inputs)))
-        return loss_keys
+        self.trace_inputs = dict()
+        traces = [trace for trace in self.traces]
+        loss_keys = self.network.get_loss_keys()
+        monitor_names = set(filter(None, to_list(self.monitor_names))).union(loss_keys)
+        traces.insert(0, TrainEssential(monitor_names=monitor_names))
+        modes = self.pipeline.get_modes() - {"test"}
```
TortoiseHam: why are we removing 'test' from modes?

fastestimator/estimator.py
```diff
@@ -64,92 +64,104 @@ def __init__(self,
...
-        loss_keys = set()
-        for op in self.network.ops:
-            if isinstance(op, UpdateOp):
-                loss_keys = loss_keys.union(set(to_list(op.inputs)))
-        return loss_keys
+        self.trace_inputs = dict()
+        traces = [trace for trace in self.traces]
+        loss_keys = self.network.get_loss_keys()
+        monitor_names = set(filter(None, to_list(self.monitor_names))).union(loss_keys)
+        traces.insert(0, TrainEssential(monitor_names=monitor_names))
```
TortoiseHam: user could instantiate without train right? if they're just inferencing?

fastestimator/estimator.py
```diff
@@ -64,92 +64,104 @@ def __init__(self,
...
+        if "eval" in modes:
+            traces.append(EvalEssential(loss_keys=loss_keys))
+        for mode in modes:
+            trace_inputs = set()
+            for trace in traces:
+                if trace.mode is None or mode in to_list(trace.mode):
+                    trace_inputs.update(filter(None, to_list(trace.inputs)))
+                    monitor_names.update(filter(None, to_list(trace.log_names)))
+            self.trace_inputs[mode] = trace_inputs
+        traces.append(Logger(log_names=monitor_names, loss_names=loss_keys))
```
TortoiseHam: this is fine for this PR, but in future should only add logger if user didn't provide one

fastestimator/estimator.py
```diff
@@ -64,92 +64,104 @@ def __init__(self,
...
+        return traces
 
     def _start(self):
         self._run_traces_on_begin()
         for self.system.epoch_idx in range(self.epochs):
```
TortoiseHam: should self.epochs be self.system.epochs ?

fastestimator/estimator.py
```diff
@@ -64,92 +64,104 @@ def __init__(self,
...
-                batch = torch_to_tf(batch)  # TODO - this should maybe be handled somewhere else...
+        self.network.load_epoch(mode=self.system.mode, epoch=self.system.epoch_idx)
+        self.system.loader = self.pipeline.get_loader(mode=self.system.mode, epoch=self.system.epoch_idx)
+        for self.system.batch_idx, batch in enumerate(self.system.loader):
             if self.system.batch_idx == self.steps_per_epoch and self.system.mode == "train":
                 break
+            batch = self._configure_tensor(batch)
             self._run_traces_on_batch_begin()
-            prediction = self.network.run_step(batch, {"mode": self.system.mode, "epoch": self.system.epoch_idx})
+            prediction = self.network.run_step(batch, {"mode": self.system.mode, "warmup": False})
```
TortoiseHam: note to self: look at these args on second pass (why dict?)

fastestimator/estimator.py
```diff
@@ -64,92 +64,104 @@ def __init__(self,
...
             self._run_traces_on_batch_end(batch, prediction)
             if self.system.mode == "train":
                 self.system.update_global_step()
+        self.network.unload_epoch()
         self._run_traces_on_epoch_end()
 
+    def _configure_tensor(self, batch):
+        if isinstance(self.system.loader, tf.data.Dataset):
```
TortoiseHam: shouldn't self.system.loader be type TFPipeline rather than exposing the tf.data.Dataset?

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+        self._verify_inputs()
+        self.effective_inputs = dict()
+        self.effective_outputs = dict()
+        self.epoch_ops = []
+
+    def _verify_inputs(self):
         for op in self.ops:
-            self.op_outputs = self.op_outputs.union(set(filter(None, to_list(op.outputs))))
-            self.op_inputs = self.op_inputs.union(set(filter(None, to_list(op.inputs))))
+            if isinstance(op, (RepeatScheduler, EpochScheduler)):
```
TortoiseHam: what about functional scheduler?

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+            else:
+                assert isinstance(op, TensorOp), "unsupported op format, must provide TensorOp in Network"
 
-    def _check_device(self):
-        if self.framework == "pytorch":
-            self.device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
-            if self.device.type == "cuda":
-                for model in self.model_list:
-                    model.to(self.device)
+    def load_epoch(self, mode, epoch):
```
TortoiseHam: typing

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+        self.epoch_ops = get_current_ops(self.ops, mode, epoch)
 
-    def prepare(self):
+    def unload_epoch(self):
         pass
 
+    def get_loss_keys(self) -> Set[str]:
+        loss_keys = set()
+        for op in self.ops:
+            if isinstance(op, (RepeatScheduler, EpochScheduler)):
```
TortoiseHam: no functional scheduler here?

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+                for epoch_op in op.get_items():
+                    if isinstance(epoch_op, UpdateOp):
+                        loss_keys.update(to_list(epoch_op.inputs))
+            else:
+                if isinstance(op, UpdateOp):
+                    loss_keys.update(to_list(op.inputs))
+        return loss_keys
+
+    def get_effective_input_keys(self, mode, total_epoches) -> Set[str]:
```
TortoiseHam: typing

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+        input_keys = set()
+        produced_keys = set()
+        for epoch in self.get_signature_epoches(total_epoches):
+            for op in get_current_ops(self.ops, mode, epoch):
+                input_keys.update(set(key for key in to_list(op.inputs) if key not in produced_keys))
+                produced_keys.update(to_list(op.outputs))
+        input_keys -= {None}
+        return input_keys
+
+    def get_all_output_keys(self, mode, total_epoches) -> Set[str]:
```
TortoiseHam: typing

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+        output_keys = set()
+        for epoch in self.get_signature_epoches(total_epoches):
+            for op in get_current_ops(self.ops, mode, epoch):
+                output_keys.update(to_list(op.outputs))
+        output_keys -= {None}
+        return output_keys
+
+    def get_signature_epoches(self, total_epoches):
```
TortoiseHam: typing

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+        signature_epoches = {0}
+        epoch_keys = {0}
+        repeat_cycles = {1}
+        for x in self.ops:
+            if isinstance(x, EpochScheduler):
+                epoch_keys.update(x.epoch_dict.keys())
+            elif isinstance(x, RepeatScheduler):
+                repeat_cycles.add(x.cycle_length)
```
TortoiseHam: rather than accessing member variables like cycle_length and epoch_dict, scheduler should perhaps have a function that takes the total epochs and returns the list of signature epochs

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+        for idx, epoch in enumerate(epoch_keys):
+            if idx + 1 < len(epoch_keys):
+                signature_epoches.update(range(epoch, epoch + min(epoch_keys[idx + 1] - epoch, least_common_cycle)))
+            else:
+                signature_epoches.update(range(epoch, epoch + least_common_cycle))
+        signature_epoches = set(epoch for epoch in signature_epoches if epoch < total_epoches)
+        return signature_epoches
+
+    @staticmethod
+    def _forward_batch(batch: Mapping[str, Any], state: Dict[str, Any], ops: List[TensorOp]):
```
TortoiseHam: any advantage / disadvantage to using System instead of state here?

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+            models.add(op.model)
+    assert models, "cannot find model in Network ops"
+
+    framework = set()
+    for model in models:
+        if isinstance(model, tf.keras.Model):
+            framework.add("tensorflow")
+        elif isinstance(model, torch.nn.Module):
+            framework.add("pytorch")
+    assert len(framework) == 1, "please make sure either tensorflow or torch model is used in network"
```
TortoiseHam: else: framework.add("unknown")

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+    return network
+
+
+class TorchNetwork(BaseNetwork):
+    def __init__(self, ops):
+        super().__init__(ops)
+        self.device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+        self.epoch_models = set()
+
+    def load_epoch(self, mode, epoch):
```
TortoiseHam: typing

fastestimator/op/op.py
```diff
@@ -54,13 +57,12 @@ def forward(self, data: Union[np.ndarray, List[np.ndarray], str, List[str]],
...
+def get_current_ops(ops: Iterable[OpType], mode: str, epoch: int = 0) -> List[OpType]:
     selected_ops = []
     for op in ops:
-        op_mode = op.mode
-        if not isinstance(op_mode, list):
-            op_mode = [op_mode]
-        if None in op_mode or mode in op_mode:
+        if isinstance(op, Scheduler):
+            op = op.get_current_value(epoch)
+        if op and (op.mode is None or mode in to_list(op.mode)):
```
TortoiseHam: why to_list(op.mode) here?

fastestimator/schedule/schedule.py
```diff
@@ -0,0 +1,74 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+
+class Scheduler:
+    def get_current_value(self, epoch: int):
```
TortoiseHam: typing, should probably raise NotImplementedError

fastestimator/schedule/schedule.py
```diff
@@ -0,0 +1,74 @@
...
+        return None
+
+    def get_items(self):
+        raise NotImplementedError
+
+
+class RepeatScheduler(Scheduler):
+    def __init__(self, repeat_list):
```
TortoiseHam: typing

fastestimator/schedule/schedule.py
```diff
@@ -0,0 +1,74 @@
...
+        assert self.cycle_length > 1, "list length must be greater than 1"
+
+    def get_current_value(self, epoch: int):
+        return self.repeat_list[epoch % self.cycle_length]
+
+    def get_items(self):
+        return self.repeat_list
+
+
+class FunctionScheduler(Scheduler):
```
TortoiseHam: this needs to implement get_items, or else the function needs to be removed from the base class

fastestimator/schedule/schedule.py
```diff
@@ -0,0 +1,74 @@
...
+    def get_current_value(self, epoch: int):
+        return self.schedule_fun(epoch)
+
+
+class EpochScheduler(Scheduler):
+    def __init__(self, epoch_dict):
+        assert isinstance(epoch_dict, dict), "must provide dictionary as epoch_dict"
+        self.epoch_dict = epoch_dict
+        self.keys = sorted(self.epoch_dict)
+        assert 0 in self.epoch_dict, "epoch 0 is missing in dictionary, use None if no op is needed"
```
TortoiseHam: why are we enforcing this rather than just adding it implicitly ourselves like before?

fastestimator/trace/trace.py
```diff
@@ -78,31 +73,26 @@ class TrainEssential(Trace):
...
         self.time_start = time.perf_counter()
 
     def on_batch_end(self, data):
-        self.num_example += self.system.batch_size
         if self.system.global_step % self.system.log_steps == 0:
             for idx, key in enumerate(self.monitor_names):
                 self.system.add_buffer(key, data[idx])
             self.elapse_times.append(time.perf_counter() - self.time_start)
-            self.system.add_buffer("examples/sec", round(self.num_example / np.sum(self.elapse_times), 1))
-            self.system.add_buffer("progress", "{:.1%}".format(self.system.global_step / self.system.total_steps))
```
TortoiseHam: we can still have progress based on epochs right? or estimate this value based on previous epochs?

fastestimator/util/util.py
```diff
@@ -76,3 +78,10 @@ def get_num_devices() -> int:
     except:
         num_devices = 1
     return num_devices
+
+
+def lcms(*numbers):
+    def lcm(a, b):
```
TortoiseHam: why hide this func def inside the parent func def?

setup.py
```diff
@@ -35,29 +35,22 @@ def get_name():
...
-    install_requires=[
-        'numpy',
-        'tensorflow>=2.1',
-        'torch @ https://download.pytorch.org/whl/torch_stable.html',
-        'albumentations',
-        'torch',  # List a second time to make the IDE happy
-        'opencv-python',
-        'scipy'
-    ],
+    install_requires=['numpy', 'albumentations', 'pyfiglet', 'opencv-python', 'scipy'],
```
TortoiseHam: still need albumentations

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
-    return sorted(signature_epochs)
 
+class Pipeline:
+    def __init__(self,
+                 train_data: Union[Dataset, DataLoader, Scheduler, tf.data.Dataset, None] = None,
+                 eval_data: Union[Dataset, DataLoader, tf.data.Dataset, None] = None,
+                 test_data: Union[Dataset, DataLoader, tf.data.Dataset, None] = None,
+                 batch_size: Union[int, Scheduler, None] = None,
+                 ops: Union[None, NumpyOp, Scheduler, List[Union[NumpyOp, Scheduler]]] = None,
+                 num_process: Union[int, None] = None):
```
TortoiseHam: this can just be 'Optional[int]' instead of Union[int, None]

fastestimator/trace/trace.py
```diff
@@ -47,26 +47,21 @@ def __init__(self,
     def on_begin(self):
         """Runs once at the beginning of training
         """
-
     def on_epoch_begin(self):
         """Runs at the beginning of each epoch
         """
-
     def on_batch_begin(self):
```
TortoiseHam: 1) trace might take inputs generated by other traces
       - Lack of output spec prevents trace sorting / true warmup validation
       - This would prevent get_inputs_by_key from working
2)  trace may want to modify batch data, impossible since no data in on_batch_begin
       - RL functionality will probably rely on this
3) Should system be passed as function arg rather than being an instance variable? alternatively, should we make it a class variable rather than an instance variable? Potential problem could then occur in jupiter if user has multiple estimators at once though

fastestimator/estimator.py
```diff
@@ -194,23 +206,21 @@ class System:
     def __init__(self,
```
TortoiseHam: System class should move to its own file, possibly inside Util folder since we don't want pipeline and/or network to import from estimator

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
-            self.datasets["eval"] = self._build_dataset(eval_data, "eval")
-            self.dataloaders["eval"] = self._build_loader(self.datasets["eval"],
-                                                          shuffle=shuffle_eval,
-                                                          num_process=num_process)
-        # All output keys from the dataset, plus any outputs from the ops
-        self.all_output_keys = {key for dataset in self.datasets.values() for key in dataset[0].keys()}
-        for op in self.ops:
-            if isinstance(op, Scheduler):
-                self.all_output_keys.update(map(lambda x: to_list(x.outputs), op.epoch_dict.values()))
+    def get_signature_epoches(self, total_epoches):
```
TortoiseHam: "epoches"?

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+                 batch_size: Union[int, Scheduler, None] = None,
+                 ops: Union[None, NumpyOp, Scheduler, List[Union[NumpyOp, Scheduler]]] = None,
+                 num_process: Union[int, None] = None):
+        self.data = {x: y for (x, y) in zip(["train", "eval", "test"], [train_data, eval_data, test_data]) if y}
+        self.batch_size = batch_size
+        self.ops = ops
+        self.num_process = num_process
+        self.fe_dataset_exist = False
+        self._verify_inputs()
+        self.ops = [] if ops is None else to_list(ops)
```
TortoiseHam: is there a more cunning way to do this rather than assigning this variable twice?

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+    def _verify_dataset(self, mode, dataset):
+        if isinstance(dataset, Dataset):
+            self.fe_dataset_exist = True
+            #batch_size check
+            assert isinstance(self.batch_size, (RepeatScheduler, EpochScheduler, int)), "unsupported batch_size format"
+            if isinstance(self.batch_size, Scheduler):
+                for batch_size in self.batch_size.get_items():
+                    assert isinstance(batch_size, int), "unsupported batch_size format"
+            #ops check
+            assert isinstance(self.ops, list), "unsupported ops format"
```
TortoiseHam: we're not auto-wrapping single ops into list anymore? (wouldn't have been a problem if self.ops was assigned properly before _verify_inputs)

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+
+    def get_loader(self, mode: str, epoch: int = 0):
+        data = self.data[mode]
+        if isinstance(data, Scheduler):
+            data = data.get_current_value(epoch)
+        if isinstance(data, Dataset):
+            op_dataset = OpDataset(data, get_current_ops(self.ops, mode, epoch), mode)
+            batch_size = self.batch_size
+            if isinstance(batch_size, Scheduler):
+                batch_size = batch_size.get_current_value(epoch)
```
TortoiseHam: do we want to support mode-based batch_size?

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+            data = DataLoader(op_dataset, batch_size=batch_size, shuffle=mode == "train", num_workers=self.num_process)
```
TortoiseHam: drop_last and shuffle should be user overridable, worker_init_fn should be lambda _: np.random.seed()

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
-
-    def show_results(self, mode: str = "train", num_steps: int = 1, epoch: int = 0):
-        data = []
-        itr = self.get_iterator(mode=mode, epoch=epoch)
-        for _ in range(num_steps):
-            data.append(next(itr))
+            if idx == num_steps:
+                break
+
+    def get_loader(self, mode: str, epoch: int = 0):
```
TortoiseHam: typing, we should probably return an iterator rater than the loader since the loaders will be of different types

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+                 eval_data: Union[Dataset, DataLoader, tf.data.Dataset, None] = None,
+                 test_data: Union[Dataset, DataLoader, tf.data.Dataset, None] = None,
+                 batch_size: Union[int, Scheduler, None] = None,
+                 ops: Union[None, NumpyOp, Scheduler, List[Union[NumpyOp, Scheduler]]] = None,
+                 num_process: Union[int, None] = None):
+        self.data = {x: y for (x, y) in zip(["train", "eval", "test"], [train_data, eval_data, test_data]) if y}
+        self.batch_size = batch_size
+        self.ops = ops
+        self.num_process = num_process
+        self.fe_dataset_exist = False
```
TortoiseHam: since this is only used in validation step let's not make it a member variable

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
-
-    def get_num_examples(self, mode: str, epoch: int) -> int:
-        raise NotImplementedError()
-
-    def get_num_steps(self, mode: str, epoch: int) -> int:
-        return self.get_num_examples(mode=mode, epoch=epoch) // self.get_batch_size(mode=mode, epoch=0)
-
-    def get_iterator(self, mode: str, epoch: int = 0, dataset: Optional[Dataset] = None) -> Iterator:
-        raise NotImplementedError()
+        return set(mode for mode in self.data.keys())
```
TortoiseHam: set(self.data.keys()) is sufficient

fastestimator/pipeline.py
vbvg2008: no we can't because we cannot use for loop to access iterator multiple times.

fastestimator/estimator.py
vbvg2008: I think we won't import system anywhere else.  if that's the case is it ok to put it inside estimator?

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+        for epoch in self.get_signature_epoches(total_epoches):
+            loader = self.get_loader(mode=mode, epoch=epoch)
+            data = next(iter(loader))
+            assert isinstance(data, dict), "please make sure data output format is dictionary"
+            output_keys = output_keys.union(set(data.keys()))
+            if isinstance(loader, DataLoader) and isinstance(loader.dataset, OpDataset):
+                for op in loader.dataset.ops:
+                    output_keys.update(to_list(op.outputs))
+        output_keys -= {None}
+        return output_keys
```
TortoiseHam: IDE thinks that output_keys will be a set[list] here instead of the desired set[str]. need to look into why

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
-    signature_epochs = {0}
-    if isinstance(val, Scheduler):
-        signature_epochs.update(val.epoch_dict.keys())
-    else:
-        for elem in val:
-            if isinstance(elem, Scheduler):
-                signature_epochs.update(elem.epoch_dict.keys())
-    return sorted(signature_epochs)
 
+class Pipeline:
```
TortoiseHam: are we giving up on the show_results() method?

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+            assert not self.num_process, "only support num_process with built-in dataset in Pipeline"
+
+    def _verify_dataset(self, mode, dataset):
+        if isinstance(dataset, Dataset):
+            self.fe_dataset_exist = True
+            #batch_size check
+            assert isinstance(self.batch_size, (RepeatScheduler, EpochScheduler, int)), "unsupported batch_size format"
+            if isinstance(self.batch_size, Scheduler):
+                for batch_size in self.batch_size.get_items():
+                    assert isinstance(batch_size, int), "unsupported batch_size format"
```
TortoiseHam: assert 0 in batch_size.epoch_dict.keys(), "Batch size must be specified for epoch 0"
            assert all(map(lambda x: x is not None, batch_size.epoch_dict.values())), "Batch size must never be None"

fastestimator/pipeline.py
TortoiseHam: we should keep the get_num_examples() call to compute the progress percentage

fastestimator/pipeline.py
```diff
@@ -13,442 +13,132 @@
...
+            #ops check
```
TortoiseHam: run pep8 formatting on file

fastestimator/pipeline.py
TortoiseHam: do we really want to initialize new datasets / dataloaders every epoch? this seems like it could add noticeable overhead, esp. since data loader owns a thread pool

fastestimator/dataset/fe_dataset.py
```diff
@@ -0,0 +1,38 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+from torch.utils.data import Dataset
+
+from fastestimator.op import get_inputs_by_op, write_outputs_by_key
+
+
+class OpDataset(Dataset):
```
TortoiseHam: lets rename this file op_dataset

fastestimator/pipeline.py
TortoiseHam: need to edit to_list in util.py to be:

def to_list(data: Any) -> List[Any]:

fastestimator/estimator.py
TortoiseHam: nvm, i see why there's discrepancy

fastestimator/estimator.py
TortoiseHam: after looking things over, not clear to me why we don't just put system inside network rather than these dictionary args. is it to do with tf.function problems?

fastestimator/estimator.py
```diff
@@ -194,23 +206,21 @@ class System:
...
-        self.batch_size = batch_size
         self.num_devices = num_devices
         self.log_steps = log_steps
         self.total_epochs = total_epochs
-        self.total_steps = total_steps
         self.epoch_idx = epoch_idx
         self.batch_idx = batch_idx
         self.buffer = {}
+        self.loader = None
+        self.network = None
```
TortoiseHam: why are we putting a reference to loader/network inside of System? this will create some circular dependency issues if we move system class to util folder

fastestimator/network.py
```diff
@@ -13,51 +13,141 @@
...
+    def _forward_batch(batch: Mapping[str, Any], state: Dict[str, Any], ops: List[TensorOp]):
+        data = None
+        for op in ops:
+            data = get_inputs_by_op(op, batch, data)
+            data = op.forward(data, state)
+            if op.outputs:
+                write_outputs_by_key(batch, data, op.outputs)
+
+
+def Network(ops):
```
TortoiseHam: # noinspection PyPep8Naming

fastestimator/estimator.py
TortoiseHam: i think it will probably play a role in the Summary object when we bring that functionality back

fastestimator/pipeline.py
TortoiseHam: not sure what you mean, `for elem in iter` is pretty standard...

fastestimator/estimator.py
TortoiseHam: even if we don't though i think it makes more sense to move it from an organization perspective

fastestimator/estimator.py
vbvg2008: because `Trace` is not designed for test or inference.

fastestimator/estimator.py
vbvg2008: We will add inferencing API later, which involves having `transform` in Network class.

here's the way to inference in the future:
```python
pipeline = ....
network = ...

#start inferencing
loader = pipeline.get_loader(mode="test")
for test_data in loader:
  result = network.transform(test_data, mode="test")
```

fastestimator/estimator.py
vbvg2008: k

fastestimator/dataset/fe_dataset.py
vbvg2008: added typing

fastestimator/estimator.py
vbvg2008: they are the same number.   but I did change the variable name from self.system.total_epochs to self.system.epochs

fastestimator/estimator.py
vbvg2008: yes, tf.function has issue accessing class variable

fastestimator/estimator.py
vbvg2008: There's no TFPipeline anymore.   

the loader can be either tf.data.dataset or pytorch loader

fastestimator/network.py
vbvg2008: `FunctionalScheduler` is designed for returning numerical values only.    it will be used for learning rate scheduler for example.

fastestimator/network.py
vbvg2008: yes sir

fastestimator/network.py
vbvg2008: `FunctionalScheduler` is designed for returning numerical values only. it will be used for learning rate scheduler for example.

fastestimator/network.py
vbvg2008: yep

fastestimator/network.py
vbvg2008: k

fastestimator/network.py
vbvg2008: nice catch

fastestimator/network.py
vbvg2008: When `RepeatScheduler` and `EpochScheduler` are used together, `RepeatScheduelr` 's signature epoch is dependent upon `EpochScheduler`. 

Therefore it doesn't make sense to have that function.

fastestimator/network.py
vbvg2008: cus tf.function doesn't like object.

fastestimator/network.py
vbvg2008: good point!

fastestimator/network.py
vbvg2008: k

fastestimator/op/op.py
vbvg2008: you are right, it's not necessary

fastestimator/schedule/schedule.py
vbvg2008: yep

fastestimator/schedule/schedule.py
vbvg2008: k

fastestimator/schedule/schedule.py
vbvg2008: yep I removed from base class

fastestimator/schedule/schedule.py
vbvg2008: to make life easier and more transparent.   it will also avoid making mistakes

fastestimator/trace/trace.py
vbvg2008: it doesn't look good to print the same message many times within one epoch like:

progress: 1/10
progress: 1/10
progress: 1/10
progress: 1/10
progress: 1/10
progress: 1/10
progress: 1/10
progress: 1/10

fastestimator/util/util.py
vbvg2008: `lcm` is only intended to be used by `lcms` for recursion purpose.   The user only need to use the `lcms`

setup.py
vbvg2008: albumentations is the second element of the list

fastestimator/pipeline.py
vbvg2008: sure, but I never know their differences.

fastestimator/estimator.py
vbvg2008: sure

fastestimator/pipeline.py
vbvg2008: nice catch!

fastestimator/pipeline.py
vbvg2008: If we still want to check for incorrect usage of pipeline, I can't think of another way right now.

fastestimator/pipeline.py
vbvg2008: you are right, the assertion is not necessary, removed it.

fastestimator/pipeline.py
vbvg2008: At least not now. We will see about that usecase in the future.

fastestimator/pipeline.py
vbvg2008: Keeping the `Pipeline` API simple is our primary goal here.  
If user wants more complicated stuff, they are sophisticated enough create a pytorch dataloader themselves and use it in `Pipeline`

fastestimator/pipeline.py
vbvg2008: try the following code and you will see:

```python
a = [1, 2, 3, 4]
it = iter(a)

#first time
for ele in it:
  print(ele)

#second time
for ele in it:
  print(ele)

```

fastestimator/pipeline.py
vbvg2008: you mean `self.fe_dataset_exist` ?   it has to be member variable because it is needed on scenarios like: `train_data` is data loader, `eval_data` is dataset.

fastestimator/pipeline.py
vbvg2008: nice!

fastestimator/pipeline.py

fastestimator/pipeline.py
vbvg2008: it's pretty convenient to show results now, first get the loader, then loop it. 

but we might add API if it is indeed needed

fastestimator/pipeline.py
vbvg2008: The batch size can also be a `RepeatScheduler`, which means it won't have `epoch_dict` variable.

Given that 0 epoch is enforced in `EpochScheduelr`, the first assertion is not necessary.
If the batch_size is indeed a scheduler, we are checking all of them to be integer, which is better than checking if None. 

Therefore, I think we should keep this code as-is.

fastestimator/pipeline.py
vbvg2008: The number of examples is complicating lots of things:
1. having number of examples won't allow us to calculate the progress, we also need the batch_size. 
2. calculating number of examples with tf.data.dataset doesn't work on all situations, then we have to accommodate different special cases.  


Users do have the epoch number and step number as a indicator of progress.

fastestimator/pipeline.py
vbvg2008: it is already, on my side.

fastestimator/pipeline.py
vbvg2008: the overhead is negligible comparing to the amount of training time needed for each epoch.   The advantage for doing this is:
* simplicity on code base
* notebook friendly.

fastestimator/dataset/fe_dataset.py
vbvg2008: done

fastestimator/estimator.py
vbvg2008: moved system.

fastestimator/network.py
vbvg2008: it doesn't complain on my side, but yeah I added that line

fastestimator/dataset/fe_dataset.py
TortoiseHam: ops should be an Optional[List[NumpyOp]] right?

fastestimator/estimator.py
TortoiseHam: XAI traces would want to run there still

fastestimator/estimator.py
TortoiseHam: hmm. that will be cool, I think that it still makes sense to wrap the insert here in an `if` but fine for now I guess.

fastestimator/estimator.py
TortoiseHam: I guess the point of this question was more: why are we storing self.epochs instead of exclusively relying on self.system.epochs?

fastestimator/network.py
TortoiseHam: you'd still reconcile the signature epochs at the top level. or are you saying that we can nest schedulers inside each other?

fastestimator/util/util.py
TortoiseHam: that doesn't seem like a particularly good reason to do this. Do you know if there are any performance ramifications of defining functions inside other functions?

fastestimator/network.py
TortoiseHam: interesting. I think there's a way we can indicate that with type hinting, though I suppose that can wait

fastestimator/trace/trace.py
TortoiseHam: right, if we did epoch based it would have to be just at epoch end I suppose

setup.py
TortoiseHam: ah, right. apparently I just can't read today

fastestimator/pipeline.py
TortoiseHam: what tf.data.dataset situations did you find that didn't work?

fastestimator/pipeline.py
TortoiseHam: good point

fastestimator/pipeline.py
TortoiseHam: we have different pep8 apparently. we'll have to figure out whose settings are actually correct

fastestimator/pipeline.py
TortoiseHam: don't we invoke get_loader once per epoch though? why is that a use case for us?

fastestimator/dataset/fe_dataset.py
vbvg2008: nope, the list will not be optional. it's an empty list if there's no ops

fastestimator/pipeline.py
TortoiseHam: worker_init_fn needs to be set regardless. 

having drop_last always on and shuffle eval always false means that some eval data will always be ignored

fastestimator/estimator.py
vbvg2008: Inferencing should not go through Estimator, therefore not through trace.

fastestimator/estimator.py
vbvg2008: the system is instantiated after we call the fit, but we need self.epochs to store the init argument.  there's absolutely no differences between either case

fastestimator/dataset/fe_dataset.py
TortoiseHam: ah, I misread something later on. In that case List[NumpyOp] rather than just list

fastestimator/network.py
vbvg2008: here's an example:
when repeat_list is [a, b] and epoch_dict is {0: c, 5:d}, when combining both scheduler, the signature epochs are: {0, 1, 5, 6}

Now question: what should be the output of repeatScheduler's signature epoch?

fastestimator/estimator.py
TortoiseHam: That seems like we're shooting ourselves in our foot. traces might be useful

fastestimator/util/util.py
vbvg2008: it can keep the lcm private such that user won't use the wrong function.

fastestimator/estimator.py
TortoiseHam: storing same info in multiple places can give rise to inconsistent states. better to have single source of truth when possible. maybe not possible here, but we could perhaps partially instantiate system on estimator creation

fastestimator/trace/trace.py
vbvg2008: 1. Trace won't write data to batch data, but instead Trace will use self.system.buffer to communicate

2. We won't allow trace to modify batch data, because there's no way to do warm up on it.  We don't have to worry about RL now.

3. I didn't get the issue. each estimator instance will have their own system instance, just like each model instance has their own variables and optimizer.

fastestimator/network.py
TortoiseHam: ah, I see

fastestimator/pipeline.py
vbvg2008: there are other parts of the code that performs some instance check on the loader.  we still can't return iterator.

fastestimator/pipeline.py
vbvg2008: when ds = ds.repeat()

when we call ds.from_generator

fastestimator/pipeline.py
vbvg2008: yep.

fastestimator/dataset/fe_dataset.py
vbvg2008: done

fastestimator/estimator.py
vbvg2008: We will think about it when we implement the inferencing API :)

fastestimator/estimator.py
vbvg2008: The system has to re-initiate itself or reset itself after every fit call.  - this is critical for notebook usage.  Therefore we might as well instantiate new system for every fit.

fastestimator/estimator.py
TortoiseHam: question about why network is needed in system still open

fastestimator/pipeline.py
vbvg2008: I think the default for dataloader is drop last = False, and we are using the default here.    no data will be lost.

fastestimator/estimator.py
vbvg2008: in case user wants to access some network property in Trace.

the goal is to enable the access and it's up to the user to use it.

fastestimator/trace/trace.py
TortoiseHam: 1. many of the current traces (tensor board, XAI as examples) link outputs from one trace to inputs of another. current system allows validation of that linking but in new system with no trace outputs it's not possible to validate

2. fine, though I anticipate that we will regret this decision in the future. we might want to put a placeholder variable in the method signature so that we can implement something later without invalidating all user traces

3. technically each trace doesn't have their own system but all references to a single system instance. I'm just worried about so many references sitting around for the whole training time but I guess it's fine

fastestimator/pipeline.py
TortoiseHam: ah, ok. still need the worker_init_fn though

fastestimator/trace/trace.py
vbvg2008: yeah I agree with your point1, we can discuss how to better approach it after merging this PR :)

fastestimator/pipeline.py
vbvg2008: trust me, more than 99% of user won't use that argument.  let's leave it for now.

TortoiseHam: no, not an input argument. It needs to be set to lambda _: np.random.seed()

TortoiseHam: even before this pr I don't think we were letting the user choose that one manually

vbvg2008: ok, added the worker_init in DataLoader call

PR #440, opened 2020-02-12
cleaning

TortoiseHam:
fix log_steps, misc cleanup, start of work towards allowing op outputs to chain into later op inputs

### DIFF
--- a/fastestimator/dataset/csv_dataset.py
+++ b/fastestimator/dataset/csv_dataset.py
@@ -1,0 +1,41 @@
+# Copyright 2019 The FastEstimator Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import os
+from typing import Dict
+
+import pandas as pd
+from torch.utils.data import Dataset
+
+
+class CSVDataset(Dataset):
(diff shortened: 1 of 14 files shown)


fastestimator/estimator.py
```diff
@@ -115,26 +104,26 @@ def _prepare_traces(self):
         traces.insert(0, TrainEssential(monitor_names=monitor_names))
         modes = self.pipeline.get_modes() - {"test"}
         if "eval" in modes:
-            traces.append(EvalEssential(loss_keys=loss_keys))
+            traces.insert(1, EvalEssential(loss_keys=loss_keys))
```
vbvg2008: why insert instead of append?

fastestimator/dataset/csv_dataset.py
```diff
@@ -0,0 +1,66 @@
...
+        self.parent_path = os.path.dirname(csv_path)
+
+    def __len__(self) -> int:
+        return len(self.data)
+
+    def __getitem__(self, index: int) -> Dict:
+        return self.data[index]
+
+
+class CSVDatasets:
```
vbvg2008: We shouldn't have two CSVDataset API

fastestimator/trace/trace.py
```diff
@@ -13,36 +13,40 @@
...
     """
+    system: System
+
     def __init__(self,
                  inputs: Union[None, str, Iterable[str], Callable] = None,
-                 mode: Union[None, str, Iterable[str]] = None,
-                 log_names=None):
+                 outputs: Union[None, str, Iterable[str]] = None,
+                 mode: Union[None, str, Iterable[str]] = None):
         if isinstance(inputs, Iterable) and not isinstance(inputs, str):
```
vbvg2008: why not just call `to_list`

fastestimator/estimator.py
TortoiseHam: eval essential before all other user provided traces in case those traces want to look at the eval essential output

fastestimator/trace/trace.py
TortoiseHam: cause we want single inputs to not be wrapped in list right?

fastestimator/dataset/csv_dataset.py
TortoiseHam: we don't, we have one CSVDataset API and one CSVDatasets API :)

PR #446, opened 2020-02-14
update trace API

TortoiseHam:
return of the ChainMap

### DIFF
--- a/fastestimator/estimator.py
+++ b/fastestimator/estimator.py
@@ -13,15 +13,15 @@
 # limitations under the License.
 # ==============================================================================
 from collections import ChainMap
-from typing import Iterable, List, Optional, Union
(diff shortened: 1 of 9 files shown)


fastestimator/estimator.py
```diff
@@ -52,14 +53,14 @@ def __init__(self,
                  steps_per_epoch: Optional[int] = None,
                  traces: Union[Trace, Iterable[Trace]] = None,
                  log_steps: Optional[int] = 100,
-                 monitor_names: Optional[str] = None):
+                 monitor_names: Optional[Union[str, Set[str]]] = None):
```
vbvg2008: The type should be iterable because we should accept list as well.

we can convert it to set later.

TortoiseHam: Done

PR #448, opened 2020-02-16
fix more memory leaks, improved pipeline training speed,  fixed potential error from tf.data

vbvg2008:


fastestimator/estimator.py
```diff
@@ -65,6 +66,7 @@ def __init__(self,
         self.trace_inputs = dict()
 
     def fit(self):
+        draw()
```
TortoiseHam: this call is in the logger now since we want user to pass None to disable all console output right?

vbvg2008: not this one, because this is our signature :)

fastestimator/estimator.py
```diff
@@ -144,12 +150,19 @@ def _run_epoch(self):
...
+            data_type = to_type(to_tensor(loader.dataset[0], target_type="tensorflow"))
+            new_loader = tf.data.Dataset.from_generator(lambda: loader, data_type)
+            new_loader = new_loader.prefetch(1)
+            if self.steps_per_epoch and self.system.mode == "train":
+                new_loader = new_loader.take(self.steps_per_epoch)
+        return new_loader
+
+    def _configure_tensor(self, loader, batch):
+        if isinstance(loader, tf.data.Dataset) and isinstance(self.network, TorchNetwork):
+            batch = to_tensor(batch, target_type="torch")
```
TortoiseHam: don't we still need the else condition to convert torch tensors to tf network?

PR #451, opened 2020-02-18
new trace: precision

geez0219:
1. complete the new trace: precision

### DIFF
--- a/fastestimator/trace/metric/precision.py
+++ b/fastestimator/trace/metric/precision.py
@@ -14,7 +14,7 @@
 # ==============================================================================
 import numpy as np
 from sklearn.metrics import precision_score
-from typing import Optional
+from typing import Optional, Union, List
 
 from fastestimator.backend.to_number import to_number
 from fastestimator.trace.trace import Trace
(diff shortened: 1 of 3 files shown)


fastestimator/trace/metric/precision.py
```diff
@@ -0,0 +1,110 @@
...
+        sample_weight (1d array-like, optional): Sample weight. If None, it will not apply sample weight. For more
+            details please refer to sklearn.metrics.precision_score. Defaults to None.
+        mode (str, optional): Restrict the trace to run only on given modes {'train', 'eval', 'test'}. None will always
+                    execute. Defaults to 'eval'.
+        output_name (str, optional): Name of the key to store to the state. Defaults to "precision".
+    """
+    def __init__(self,
+                 true_key: str,
+                 pred_key: str,
+                 labels=None,
```
TortoiseHam: can you include typing for all args?

fastestimator/trace/metric/precision.py
```diff
@@ -0,0 +1,110 @@
...
+from fastestimator.backend.to_number import to_number
+from fastestimator.trace.trace import Trace
+from fastestimator.util import Data
+
+
+class Precision(Trace):
+    """Computes precision for classification task and report it back to logger.
+
+    Args:
+        true_key (str): Name of the keys in the ground truth label in data pipeline.
```
TortoiseHam: we're going to type hint in code instead of doc string now

fastestimator/trace/metric/precision.py
geez0219: finish adding typing

fastestimator/trace/metric/precision.py
geez0219: removed arguments type in the docstring

PR #456, opened 2020-02-19
adding estimator.test API

vbvg2008:
fixing the typing of two traces


fastestimator/estimator.py
```diff
@@ -62,14 +62,40 @@ def __init__(self,
...
+        return self._start_test()
+
+    def _prepare_traces(self, modes: Set[str]):
+        self.trace_inputs = dict()
+        loss_keys = set()
+        traces = [trace for trace in self.traces]
+        if "train" in modes:
+            loss_keys = self.network.get_loss_keys()
+            traces.insert(0, TrainEssential())
+            if "eval" in modes:
```
TortoiseHam: what if user wants to run 'eval' without 'train' to debug something?

fastestimator/estimator.py
```diff
@@ -132,6 +140,10 @@ def _start(self):
             pass  # On early stopping we still want to run the final traces and return results
         self._run_traces_on_end()
 
+    def _start_test(self):
```
TortoiseHam: this misses run_traces_on_begin and run_traces_on_end which may be required for test traces too if setup / teardown occurs there

fastestimator/estimator.py
vbvg2008: the differentiation of `eval` and `test` are:  eval happens during training, test does not. 

Therefore, if user wants to debug, put the data as test data then call estimator.test()

fastestimator/estimator.py
vbvg2008: `eval` doesn't run it, therefore, `test` shouldn't run it

fastestimator/estimator.py
```diff
@@ -62,14 +62,40 @@ def __init__(self,
         self.monitor_names = to_set(monitor_names)
         self.system = System(log_steps=log_steps, epochs=epochs, max_steps_per_epoch=max_steps_per_epoch)
         self.trace_inputs = dict()
+        self.traces_in_use = []
```
TortoiseHam: why is this distinction necessary when we already have a system to inspect the mode of a trace?

fastestimator/trace/metric/accuracy.py
```diff
@@ -32,7 +32,11 @@ class Accuracy(Trace):
         mode: What mode to execute in (None to execute in all modes)
         output_name: What to call the output from this trace (for example in the logger output)
     """
-    def __init__(self, true_key: str, pred_key: str, mode: Optional[str] = "eval", output_name: str = "accuracy"):
+    def __init__(self,
+                 true_key: str,
+                 pred_key: str,
+                 mode: Union[str, List] = ["eval", "test"],
```
TortoiseHam: Union[str, List[str]]

fastestimator/estimator.py
vbvg2008: we need a variable to store the currently modified trace. (without overwriting the self.trace)

fastestimator/trace/metric/mcc.py
```diff
@@ -36,7 +36,11 @@ class MCC(Trace):
         mode: What mode to execute in (None to execute in all modes)
         output_name: What to call the output from this trace (for example in the logger output)
     """
-    def __init__(self, true_key: str, pred_key: str, mode: Optional[str] = "eval", output_name: str = "mcc"):
+    def __init__(self,
+                 true_key: str,
+                 pred_key: str,
+                 mode: Union[str, List] = ["eval", "test"],
```
TortoiseHam: Union[str, List[str]]

fastestimator/trace/metric/accuracy.py
vbvg2008: done

fastestimator/trace/metric/precision.py
```diff
@@ -27,37 +28,16 @@ class Precision(Trace):
...
-
-    y_true: List[int]
-    y_pred: List[int]
-
     def __init__(self,
                  true_key: str,
                  pred_key: str,
-                 labels: Optional[list] = None,
-                 pos_label: Union[int, str] = 1,
-                 average: str = 'auto',
```
TortoiseHam: why are we dropping all this functionality?

fastestimator/trace/metric/mcc.py
vbvg2008: done

fastestimator/trace/metric/precision.py
```diff
@@ -27,37 +28,16 @@ class Precision(Trace):
...
-                 sample_weight: Optional[np.ndarray] = None,
-                 mode: str = "eval",
-                 output_name: Optional[str] = "precision"):
-                 
+                 mode: Union[str, List] = ["eval", "test"],
```
TortoiseHam: Union[str, List[str]]

fastestimator/trace/metric/recall.py
```diff
@@ -28,36 +28,17 @@ class Recall(Trace):
...
-                 pred_key: Optional[str] = None,
-                 labels: Optional[str] = None,
-                 pos_label: Optional[Union[int, str]] = 1,
-                 average: Optional[str] = 'auto',
-                 sample_weight: Optional[np.ndarray] = None,
-                 mode: Optional[str] ="eval",
-                 output_name: Optional[str] = "recall"):
-
+                 pred_key: str,
+                 mode: Union[str, List] = ["eval", "test"],
```
TortoiseHam: Union[str, List[str]]

fastestimator/trace/metric/recall.py
```diff
@@ -28,36 +28,17 @@ class Recall(Trace):
...
         mode: Restrict the trace to run only on given modes {'train', 'eval', 'test'}. None will always
                     execute. Defaults to 'eval'.
         output_name: Name of the key to store to the state. Defaults to "recall".
     """
     def __init__(self,
                  true_key: str,
-                 pred_key: Optional[str] = None,
-                 labels: Optional[str] = None,
-                 pos_label: Optional[Union[int, str]] = 1,
-                 average: Optional[str] = 'auto',
```
TortoiseHam: why are we dropping all this functionality?

fastestimator/trace/metric/precision.py
vbvg2008: some doesn't apply to us, some are unnecessary.

fastestimator/util/data.py
```diff
@@ -28,4 +28,4 @@ def write_without_log(self, key: str, value: Any):
         self.maps[1][key] = value
 
     def read_logs(self, extra_keys: Set[str]) -> Dict[str, Any]:
-        return {**{k: v for k, v in self.maps[1].items() if k in extra_keys}, **self.maps[0]}
+        return {**{k: self.__getitem__(k) for k in extra_keys if self.__contains__(k)}, **self.maps[0]}
```
TortoiseHam: this will actually be slower than the other implementation with no functional benefits

fastestimator/trace/trace.py
```diff
@@ -168,10 +168,8 @@ def _print_message(self, header: str, data: Data, log_epoch: bool = False):
         log_message = header
         if log_epoch:
             log_message += "epoch: {}; ".format(self.system.epoch_idx)
-        for key, val in data.read_logs(to_set(self.inputs)).items():
+        for key, val in data.read_logs(self.inputs).items():
```
TortoiseHam: still need to_set to match read_logs type spec

vbvg2008: it's already a set

fastestimator/estimator.py
TortoiseHam: start and end are not mode dependent. eval technically does run it

fastestimator/trace/trace.py
```diff
@@ -144,10 +144,8 @@ class Logger(Trace):
     Args:
         extra_log_keys (set): set of keys to print from system buffer
     """
-    def __init__(self, extra_log_keys: Set[str], loss_names: Set[str]):
-        super().__init__(inputs=extra_log_keys | loss_names)
-        self.extra_log_keys = extra_log_keys
-        self.loss_names = loss_names
+    def __init__(self, extra_log_keys: Set[str]):
+        super().__init__(inputs=extra_log_keys)
```
TortoiseHam: what's the plan for keeping the loss keys in scope during training so they still get brought back from gpu to cpu?

fastestimator/estimator.py
TortoiseHam: there may be eval ops that they're trying to debug though, which wouldn't be caught be test mode. or are we saying that there is never a situation where network would be different for eval vs test?

fastestimator/estimator.py
TortoiseHam: why are the traces modified though?

fastestimator/trace/trace.py
TortoiseHam: trace inputs are all converted to_list in super init right?

fastestimator/util/data.py
vbvg2008: right, just realized, changed back.

fastestimator/estimator.py
vbvg2008: they can make steps_per_epoch to be 1, then run evall directly.         Running eval separately doesn't make sense.

fastestimator/trace/metric/precision.py
TortoiseHam: why would things like sample weighting be unnecessary? Those params exist because people use them...

fastestimator/estimator.py
TortoiseHam: so in pipeline are we going to assert that if eval_data is present then train_data must be non None?

vbvg2008: we don't need to assert that in pipeline.    because pipeline can be independently used for non-training purpose as well.

fastestimator/estimator.py
vbvg2008: sure. I will add it there too.

fastestimator/estimator.py
vbvg2008: originally we have self.traces = self._prepare_traces()

fastestimator/trace/metric/precision.py
vbvg2008: users are expected to provide weight of all samples together. in deep learning, it's not practical to provide that list (consider having to create a list with 1.3million element for ImageNet).

If we are going to do it in the future, we may have to read from batch dictionary instead of providing that array upfront.

fastestimator/trace/metric/precision.py
vbvg2008: done

fastestimator/trace/metric/recall.py
vbvg2008: done

fastestimator/trace/trace.py
vbvg2008: oh right!  changed it

fastestimator/trace/trace.py
vbvg2008: During training, loss keys are included in Logger's `extra_log_keys`

fastestimator/trace/metric/recall.py
vbvg2008: same as precision

fastestimator/trace/metric/precision.py
TortoiseHam: fair enough. don't we still need pos_label though? since precision is not an asymmetric metric?

fastestimator/trace/trace.py
TortoiseHam: ah ok

fastestimator/estimator.py
TortoiseHam: do we also need to call _prepare_system() and then set those variables appropriately here? what would the appropriate value for current epoch even be in this case?

fastestimator/estimator.py
```diff
@@ -56,24 +56,46 @@ def __init__(self,
                  monitor_names: Union[None, str, Iterable[str]] = None):
         self.pipeline = pipeline
         self.network = network
+        self.epochs = epochs
```
TortoiseHam: this should not be redundantly recorded, self.system is source of truth for this info

fastestimator/estimator.py
```diff
@@ -84,54 +106,39 @@ def _warmup(self):
...
                 if "eval" in self.pipeline.get_modes():
                     self.system.mode = "eval"
                     self._run_epoch()
         except EarlyStop:
             pass  # On early stopping we still want to run the final traces and return results
         self._run_traces_on_end()
 
+    def _start_test(self):
+        self.system.mode = "test"
+        self.system.epoch_idx = self.epochs
```
TortoiseHam: self.system.epoch_idx = self.system.total_epochs

fastestimator/estimator.py
TortoiseHam: i think we still need to call _prepare_system() here too - suppose user builds an estimator and calls .test() without first calling .fit(). the traces would then not have received the system variable yet right?

fastestimator/estimator.py
vbvg2008: done

fastestimator/estimator.py
vbvg2008: done

PR #457, opened 2020-02-19
update mixup_batch.py with per image beta sampling

dibya-pati:


fastestimator/op/tensorop/augmentation/mixup_batch.py
```diff
@@ -1,54 +1,66 @@
...
+        outputs: key to store the mixed-up input
+        mode: what mode to execute in. Probably 'train'
+        alpha: the alpha value defining the beta distribution to be drawn from during training
+    """
+    def __init__(self, inputs=None, outputs=None, mode=None, alpha=1.0, **kwargs):
+        assert alpha > 0, "Mixup alpha value must be greater than zero"
+        super().__init__(inputs=inputs, outputs=outputs, mode=mode)
+        self.alpha = tf.constant(alpha)
+        self.beta = tfp.distributions.Beta(alpha, alpha)
+        self.sharedbeta = kwargs['sharedbeta']
```
TortoiseHam: this should not be kwargs

PR #468, opened 2020-02-24
finished multi-gpu support for tensorflow and other clean up

vbvg2008:


fastestimator/backend/to_shape.py
```diff
@@ -0,0 +1,41 @@
...
+# limitations under the License.
+# ==============================================================================
+
+from typing import Any
+
+import numpy as np
+
+
+def to_shape(data: Any) -> Any:
+    """convert the value of any data sturcture to data type
```
TortoiseHam: what does this doc string mean?

fastestimator/backend/to_shape.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Args:
+        data (Any): source data
```
TortoiseHam: redundant doc string typing

fastestimator/backend/to_shape.py
vbvg2008: fixed

fastestimator/backend/to_shape.py
vbvg2008: nice catch

fastestimator/estimator.py
```diff
@@ -216,27 +229,59 @@ def _start_test(self):
...
     def _configure_loader(self, loader):
         new_loader = loader
-        if isinstance(loader, DataLoader) and isinstance(self.network, TFNetwork):
-            data_type = to_type(to_tensor(loader.dataset[0], target_type="tensorflow"))
-            new_loader = tf.data.Dataset.from_generator(lambda: loader, data_type)
+        if isinstance(new_loader, DataLoader) and isinstance(self.network, TFNetwork):
+            batch = to_tensor(loader.dataset[0], target_type="tensorflow")
+            data_type = to_type(batch)
+            data_shape = to_shape(batch)
+            for key, shape in data_shape.items():
```
TortoiseHam: is this reasonable for nested data?

fastestimator/network.py
```diff
@@ -40,8 +40,9 @@ def _verify_inputs(self):
             else:
                 assert isinstance(op, TensorOp), "unsupported op format, must provide TensorOp in Network"
 
-    def load_epoch(self, mode: str, epoch: int):
-        self.epoch_ops = get_current_ops(self.ops, mode, epoch)
+    def load_epoch(self, mode: str, epoch: int) -> Set[TensorOp]:
```
TortoiseHam: are epoch ops really a set? seems like conceptually this would be a list

fastestimator/estimator.py
vbvg2008: yes, the generator needs it :) 

Once we move out of the generator for multi-gpu, we can get rid of it

fastestimator/network.py
vbvg2008: nice catch!

fastestimator/network.py
```diff
@@ -143,39 +164,24 @@ def __init__(self, ops: Iterable[Union[TensorOp, Scheduler[TensorOp]]]):
         self.device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
         self.epoch_models = set()
 
-    def load_epoch(self, mode: str, epoch: int):
-        self.epoch_ops = get_current_ops(self.ops, mode, epoch)
+    def load_epoch(self, mode: str, epoch: int) -> List[TensorOp]:
+        epoch_ops = get_current_ops(self.ops, mode, epoch)
```
TortoiseHam: should self.epoch_ops be removed from the init method?

vbvg2008: 👍

fastestimator/pipeline.py
```diff
@@ -15,16 +15,17 @@
...
 
 import numpy as np
 import tensorflow as tf
-from torch.utils.data import DataLoader, Dataset
 
 from fastestimator.dataset.dataset import OpDataset
 from fastestimator.op import NumpyOp, get_current_ops
 from fastestimator.schedule import EpochScheduler, RepeatScheduler, Scheduler
-from fastestimator.util.util import lcms, to_list
+from fastestimator.util.util import Timer, lcms, to_list
```
TortoiseHam: unused import: Timer

vbvg2008: 💯

fastestimator/summary/system.py
```diff
@@ -34,8 +34,8 @@ class System:
 
     def __init__(self,
                  network: BaseNetwork,
-                 mode: str = "train",
-                 num_devices: int = get_num_devices(),
+                 mode: Optional[str] = None,
```
TortoiseHam: why is mode optional? shouldn't the system always be in one mode or another?

fastestimator/util/util.py
```diff
@@ -135,3 +177,30 @@ def strip_suffix(target: Optional[str], suffix: Optional[str]) -> Optional[str]:
     if target[-s_len:] == suffix:
         return target[:-s_len]
     return target
+
+
+def per_replica_to_global(data):
```
TortoiseHam: does this belong in util or backend?

vbvg2008: util, because it is only used by tensorflow.

backend should contain things worked for both tf and torch

TortoiseHam: also, Typing

fastestimator/summary/system.py
vbvg2008: warmup is using None as mode.  there's an if condition that needs it.

fastestimator/util/util.py
vbvg2008: 🥇

fastestimator/estimator.py
TortoiseHam: i guess my question is more: will this work for nested data? if i'm reading these 2 lines properly we're figuring out the real shape but then replacing it with a list of None ?

vbvg2008: yes it will work on nested data.

we need the shape to be None because the shape between batch may change, but the rank won't (unless that's our assumption)

fastestimator/summary/system.py
TortoiseHam: ah, ok. we should update the class-var hint then (line 23) with that explanation

fastestimator/util/util.py
TortoiseHam: k

fastestimator/estimator.py
TortoiseHam: but doesn't the following just throw out all of the actual shape information except for batch size?

data_shape[key] = [None] * (len(shape) + 1)

vbvg2008: right.  we are not using the full potential of `get_shape` function,    but I just don't want to implement a `get_rank` because it will be redundant.

TortoiseHam: what if batch is as follows:

[ {"x": {"a":[5,10,15], "b":[1,2,3,4,5]}}, ... ]

then data_shape["x"] = [None, None, None] which doesn't seem useful

vbvg2008: interesting... here we go `get_rank`

fastestimator/summary/system.py
vbvg2008: done

PR #472, opened 2020-02-26
added CSV logger trace

purujitb:


fastestimator/util/util.py
```diff
@@ -135,3 +135,19 @@ def strip_suffix(target: Optional[str], suffix: Optional[str]) -> Optional[str]:
     if target[-s_len:] == suffix:
         return target[:-s_len]
     return target
+
+def is_number(s: str):  # pylint: disable=invalid-name
```
TortoiseHam: def is_number(s: str) -> bool:

TortoiseHam: also, maybe we should rename 's' instead of just suppressing the invalid name lint

fastestimator/trace/io/csv_logger.py
```diff
@@ -0,0 +1,77 @@
...
+            will overwrite the existed file. Defaults to False.
+        mode: Restrict the trace to run only on given modes {'train', 'eval',
+            'test'}. None will always execute. Defaults to 'eval'.
+    """
+    def __init__(self,
+                 filename: str,
+                 monitor_names: Optional[List[str]] = None,
+                 separator: str = ", ",
+                 append: bool = False,
+                 mode: str = "eval"):
```
TortoiseHam: shouldn't mode default to none here? or else at least ("eval", "test"). in any case, the type should be Union[None, str, Set[str]]

fastestimator/trace/io/csv_logger.py
```diff
@@ -0,0 +1,77 @@
...
+        self.keys = monitor_names if monitor_names is None else to_list(monitor_names)
+        super().__init__(inputs="*" if self.keys is None else monitor_names, mode=mode)
+        self.separator = separator
+        self.filename = filename
+        self.append = append
+
+    def on_begin(self, data: Data):
+        self.file = open(self.filename, 'a' if self.append else 'w')
```
TortoiseHam: all member variables for classes should be declared in init method, could just set to None in this case

fastestimator/trace/io/csv_logger.py
```diff
@@ -0,0 +1,77 @@
...
+        mode: Restrict the trace to run only on given modes {'train', 'eval',
+            'test'}. None will always execute. Defaults to 'eval'.
+    """
+    def __init__(self,
+                 filename: str,
+                 monitor_names: Optional[List[str]] = None,
+                 separator: str = ", ",
+                 append: bool = False,
+                 mode: str = "eval"):
+        self.keys = monitor_names if monitor_names is None else to_list(monitor_names)
```
TortoiseHam: there should be a way to leverage self.inputs such that self.keys is not necessary

fastestimator/trace/io/csv_logger.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def on_begin(self, data: Data):
+        self.file = open(self.filename, 'a' if self.append else 'w')
+        self.file_empty = os.stat(self.filename).st_size == 0
+
+    def on_epoch_end(self, data: Data):
+        if self.keys is None:
+            self._infer_keys(data)
+        self._make_header()
+        vals = [data.get(key, "") for key in self.keys]
```
TortoiseHam: some keys (like epoch_idx) may be inside self.system rather than data

fastestimator/trace/io/csv_logger.py
```diff
@@ -0,0 +1,77 @@
...
+        vals = [str(val.numpy()) if hasattr(val, "numpy") else str(val) for val in vals]
```
TortoiseHam: why would we want to cast regular ints to string?

fastestimator/util/util.py
purujitb: Changing this

fastestimator/trace/io/csv_logger.py
purujitb: Changing this

fastestimator/trace/io/csv_logger.py
purujitb: adding initializer to constructor

fastestimator/trace/io/csv_logger.py
purujitb: this is done to write the values to the file, which needs a str

PR #473, opened 2020-02-26
Simplify CSV Logger

vbvg2008:


fastestimator/trace/io/csv_logger.py
```diff
@@ -12,66 +12,45 @@
...
-                monitored_keys.append(key)
-            elif hasattr(val, "numpy") and len(val.numpy().shape) == 1:
-                monitored_keys.append(key)
-        self.keys = sorted(monitored_keys)
-
-    def _make_header(self):
-        if self.file_empty:
-            self.file.write(self.separator.join(self.keys))
-            self.file_empty = False
+        df = pd.DataFrame(data=self.data)
```
TortoiseHam: waiting until the end to write this file may create large memory burden

fastestimator/trace/io/csv_logger.py
```diff
@@ -12,66 +12,45 @@
...
+                 mode: Union[str, Set[str]] = ("eval", "test")):
+        super().__init__(inputs=to_list(monitor_names), mode=mode)
         self.filename = filename
-        self.append = append
+        self.data = None
 
     def on_begin(self, data: Data):
-        self.file = open(self.filename, 'a' if self.append else 'w')
-        self.file_empty = os.stat(self.filename).st_size == 0
+        self.data = {key: [] for key in self.inputs}
```
TortoiseHam: make self.data a defaultdict, then don't need to do this initialization

fastestimator/trace/io/csv_logger.py
```diff
@@ -12,66 +12,45 @@
...
     def __init__(self,
                  filename: str,
-                 monitor_names: Optional[List[str]] = None,
-                 separator: str = ", ",
-                 append: bool = False,
-                 mode: str = "eval"):
-        self.keys = monitor_names if monitor_names is None else to_list(monitor_names)
-        super().__init__(inputs="*" if self.keys is None else monitor_names, mode=mode)
-        self.separator = separator
+                 monitor_names: Union[List[str], str],
```
TortoiseHam: making monitor names mandatory seems less desirable than previous version

vbvg2008: well, anyone who uses the trace must have a reason right?  then the reason is mandatory.

fastestimator/trace/io/csv_logger.py
vbvg2008: what do you mean

fastestimator/trace/io/csv_logger.py
vbvg2008: storing a couple of numbers won't hurt.   we are doing the same in every single metric trace.   if user want to write a 3D array into a CSV file, then they will have to pay the price for it lol

fastestimator/trace/io/csv_logger.py
TortoiseHam: what if the reason is I want to record everything to csv?

fastestimator/trace/io/csv_logger.py
TortoiseHam: instead of self.data = {key: [].... } you can just say self.data = defaultdict(list) in the init method. then just always invoke .append() later

fastestimator/trace/io/csv_logger.py
vbvg2008: done

fastestimator/trace/io/csv_logger.py
vbvg2008: done

fastestimator/trace/io/csv_logger.py
```diff
@@ -12,66 +12,47 @@
...
+        self.data = defaultdict(list)
 
     def on_epoch_end(self, data: Data):
-        if self.keys is None:
-            self._infer_keys(data)
-        self._make_header()
-        vals = [data.get(key, "") for key in self.keys]
-        vals = [str(val.numpy()) if hasattr(val, "numpy") else str(val) for val in vals]
-        self.file.write("\n" + self.separator.join(vals))
+        self.data["mode"].append(self.system.mode)
```
TortoiseHam: are we sure that we always want to put mode and epoch into the data? what if the user wants some other system variables?

PR #475, opened 2020-02-27
f1_score metric added

Vivek305:


TortoiseHam:
@vbvg2008 I suspect you'll want this to follow the simplified interface?

PR #477, opened 2020-02-28
1.0dev

hanskrupakar:
Backend-agnostic model saver

### DIFF
--- a/fastestimator/trace/io/model_saver.py
+++ b/fastestimator/trace/io/model_saver.py
@@ -12,10 +12,84 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import os
+from typing import Union
 
-from fastestimator.trace.trace import Trace
+import numpy as np
+import tensorflow as tf
+import torch
+
+from fastestimator.trace import Trace
+from fastestimator.util import Data
 
 
 class ModelSaver(Trace):
-    def __init__(self, model):
-        pass
+    """Save trained model in hdf5 format.
+
+    Args:
+        model: Model instance defined by user (tf.keras.Model or torch.nn.Module)
+        model_name (str): Name of FE model.
+        save_dir (str): Directory to save the trained models.
+        save_freq (int, optional): Number of epochs to save models. Cannot be used with `save_best_only=True`. Defaults
+            to 1.
+        save_best (bool, str, optional): Best model saving monitor name. If True, the model loss is used. Defaults to
+            False.
+        save_best_mode (str, optional): Can be `'min'`, `'max'`, or `'auto'`. Defaults to 'max'.
+
+    """
+    def __init__(self, model: Union[tf.keras.Model, torch.nn.Module], model_name: str,
+                 save_dir: str, save_freq: int = 1, save_best: Union[bool, str]=False, save_best_mode:str='max'):
+
+        if save_best is True:
+            save_best = 'accuracy'
+
+        if isinstance(save_best, str):
+            super().__init__(inputs=save_best, mode='eval')
+        else:
+            super().__init__(mode='train')
+
+        if isinstance(model, tf.keras.Model):
+            self.ext = '.h5'
+        else:
+            self.ext = '.pt'
+
+        self.model = model
+        self.model_name=model_name
+        self.save_dir = save_dir
+        self.save_best = save_best
+        self.save_best_mode = save_best_mode
+        self.save_freq = save_freq
+        assert isinstance(self.save_freq, int), "save_freq must be integer"
+        if self.save_best_mode == "min":
+            self.best = np.Inf
+            self.monitor_op = np.less
+        elif self.save_best_mode == "max":
+            self.best = -np.Inf
+            self.monitor_op = np.greater
+        else:
+            raise ValueError("save_best_mode must be either 'min' or 'max'")
+        self.model = model
+
+    def on_begin(self, data: Data):
+        if self.save_dir:
+            self.save_dir = os.path.normpath(self.save_dir)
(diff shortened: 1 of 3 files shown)


fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+        save_best (bool, str, optional): Best model saving monitor name. If True, the model loss is used. Defaults to
+            False.
+        save_best_mode (str, optional): Can be `'min'`, `'max'`, or `'auto'`. Defaults to 'max'.
+
+    """
+    def __init__(self, model: Union[tf.keras.Model, torch.nn.Module], model_name: str,
+                 save_dir: str, save_freq: int = 1, save_best: Union[bool, str]=False, save_best_mode:str='max'):
+
+        if save_best is True:
+            save_best = 'accuracy'
```
TortoiseHam: why would we assume 'accuracy' instead of something like loss? perhaps don't allow boolean as an argument for save_best

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+
+        if isinstance(model, tf.keras.Model):
+            self.ext = '.h5'
+        else:
+            self.ext = '.pt'
+
+        self.model = model
+        self.model_name=model_name
+        self.save_dir = save_dir
+        self.save_best = save_best
```
TortoiseHam: no need to store self.save_best since it will be in self.inputs already

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+        self.save_freq = save_freq
+        assert isinstance(self.save_freq, int), "save_freq must be integer"
+        if self.save_best_mode == "min":
+            self.best = np.Inf
+            self.monitor_op = np.less
+        elif self.save_best_mode == "max":
+            self.best = -np.Inf
+            self.monitor_op = np.greater
+        else:
+            raise ValueError("save_best_mode must be either 'min' or 'max'")
```
TortoiseHam: then why does it say we allow 'auto' in doc string?

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+        self.model = model
+
+    def on_begin(self, data: Data):
+        if self.save_dir:
```
TortoiseHam: this if check should be an assert in the init instead. no point in a model saver if there's no place to save the model

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+            self.save_dir = os.path.normpath(self.save_dir)
+            os.makedirs(self.save_dir, exist_ok=True)
+
+    def on_epoch_end(self, data: Data):
+        if isinstance(self.save_best, str):
```
TortoiseHam: this could just be 'if self.inputs' to remove the redundant member variable

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+        if isinstance(model, tf.keras.Model):
+            self.ext = '.h5'
+        else:
+            self.ext = '.pt'
+
+        self.model = model
+        self.model_name=model_name
+        self.save_dir = save_dir
+        self.save_best = save_best
+        self.save_best_mode = save_best_mode
```
TortoiseHam: don't need to store self.save_best_mode since you're generating self.monitor_op

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+        if self.save_dir:
+            self.save_dir = os.path.normpath(self.save_dir)
+            os.makedirs(self.save_dir, exist_ok=True)
+
+    def on_epoch_end(self, data: Data):
+        if isinstance(self.save_best, str):
+            if self.monitor_op(data[self.save_best], self.best):
+                self.best = data[self.save_best]
+                self._save_model("{}_best_{}{}".format(self.model_name, self.save_best, self.ext))
+                #self._save_model("{}_best_{}_{}.h5".format(self.model_name, self.save_best, str(self.best)))
```
TortoiseHam: remove dead code

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
+        elif self.system.epoch_idx % self.save_freq == 0:
+            self._save_model("{}_epoch_{}_step_{}{}".format(
+                self.model_name, self.system.epoch_idx, self.system.batch_idx, self.ext))
+
+    def _save_model(self, name: str):
+        if self.save_dir:
```
TortoiseHam: no need for this if check - assert it in init

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,84 @@
...
 
 
 class ModelSaver(Trace):
-    def __init__(self, model):
-        pass
+    """Save trained model in hdf5/pt format.
+
+    Args:
+        model: Model instance defined by user (tf.keras.Model or torch.nn.Module)
+        model_name (str): Name of FE model.
```
TortoiseHam: remove type hints from function doc

hanskrupakar:
@TortoiseHam you should really take a look at the model saver in the master branch. Most of the code patterns here are from there, I just spoofed the functionality....

Let me make some of the changes you talked about though.

TortoiseHam:
@hanskrupakar we are trying to follow better code patterns now

vbvg2008:
there are some apphub code mixed with this PR, closing this PR and I will make a new one about ModelSaver

PR #480, opened 2020-03-02
complete gan examples, revise dnn exmaples, re-design cross entropy loss, re-design modelsaver

vbvg2008:


apphub/tabular/dnn/dnn_tf.py
```diff
@@ -0,0 +1,67 @@
...
+    model = fe.build(model_fn=create_dnn, optimizer_fn="adam")
+    network = fe.Network(ops=[
+        ModelOp(inputs="x", model=model, outputs="y_pred"),
+        CrossEntropy(inputs=("y_pred", "y"), outputs="ce"),
+        UpdateOp(model=model, loss_name="ce")
+    ])
+
+    # step 3.prepare estimator
+    traces = [
+        Accuracy(true_key="y", pred_key="y_pred"),
```
TortoiseHam: in our examples we should consider using MCC instead of accuracy to push it as a best practice

fastestimator/backend/load_model.py
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+# ==============================================================================
+from typing import Union
+
+import tensorflow as tf
+import torch
+
+
+def load_model(model: Union[tf.keras.Model, torch.nn.Module], weights_path: str):
+    """Save tensorflow or pytorch model weights to a specific directory
```
TortoiseHam: does this function save weights or load weights?

vbvg2008: thank god it's not a formatting issue. lol

apphub/tabular/dnn/dnn_tf.py
vbvg2008: we can have MCC in other examples, not here tho

fastestimator/backend/sparse_categorical_crossentropy.py
```diff
@@ -0,0 +1,53 @@
...
+    Returns:
+        Tensor: categorical cross entropy
+    """
+    assert type(y_pred) == type(y_true), "y_pred and y_true must be same tensor type"
+    assert isinstance(y_pred, (tf.Tensor, torch.Tensor)), "only support tf.Tensor or torch.Tensor as y_pred"
+    assert isinstance(y_true, (tf.Tensor, torch.Tensor)), "only support tf.Tensor or torch.Tensor as y_true"
+    if isinstance(y_pred, tf.Tensor):
+        ce = tf.losses.sparse_categorical_crossentropy(y_pred=y_pred, y_true=y_true, from_logits=from_logits)
+    else:
+        y_true = y_true.view(-1)
```
TortoiseHam: just out of curiosity, why does pytorch keep these little data manipulations? did they not build their functions to handle tensors normally?

fastestimator/backend/update_model.py
```diff
@@ -29,7 +29,7 @@ def update_model(model: Union[tf.keras.Model, torch.nn.Module],
             gradients = tape.gradient(loss, model.trainable_variables)
             model.current_optimizer.apply_gradients(zip(gradients, model.trainable_variables))
     elif isinstance(model, torch.nn.Module):
-        loss.backward()
+        loss.backward(retain_graph=True)
```
TortoiseHam: if this flag needs to be set, how was our code working before?

fastestimator/backend/sparse_categorical_crossentropy.py
vbvg2008: obviously pytorch's robustness not as good as tensorflow.  everything has to be correct shape and type.

fastestimator/dataset/data/breast_cancer.py
```diff
@@ -0,0 +1,34 @@
...
+from sklearn.datasets import load_breast_cancer
+from sklearn.model_selection import train_test_split
+from sklearn.preprocessing import StandardScaler
+
+from fastestimator.dataset.numpy_dataset import NumpyDataset
+
+
+def load_data() -> Tuple[NumpyDataset, NumpyDataset]:
+    (X, y) = load_breast_cancer(True)
+    x_train, x_eval, y_train, y_eval = train_test_split(X, y, test_size=0.2)
```
TortoiseHam: should we seed this split so that users always get exactly the same train and eval sets?

fastestimator/backend/update_model.py
vbvg2008: this flag is needed if there are multiple UpdateOp in network ops.   by default, the whole graph is released when we call backward. setting it to True will prevent that from happening. 

there's no performance or memory differences from my benchmark.

fastestimator/dataset/data/breast_cancer.py
vbvg2008: I guess not, random splitting has been a convention for long.

fastestimator/network.py
```diff
@@ -145,8 +146,20 @@ def _collect_models(ops: Iterable[Union[TensorOp, Scheduler[TensorOp]]]) -> Set[
     return models
 
 
+def _associate_loss(ops: Iterable[Union[TensorOp, Scheduler[TensorOp]]]):
```
TortoiseHam: shouldn't this functionality be handled inside the model op?

fastestimator/network.py
```diff
@@ -249,29 +262,35 @@ def forward_step_static(self,
...
-    optimizer_fn = to_list(optimizer_fn)
-    assert len(models) == len(optimizer_fn)
-    for idx, (model, optimizer_def) in enumerate(zip(models, optimizer_fn)):
-        models[idx] = _fe_compile(model, optimizer_def)
+    models, optimizer_fn = to_list(model_fn()), to_list(optimizer_fn),
+    if weights_path:
+        weights_path = to_list(weights_path)
+    else:
+        weights_path = [None] * len(models)
+    assert len(models) == len(optimizer_fn) == len(weights_path)
```
TortoiseHam: helper text on assert

fastestimator/network.py
vbvg2008: interesting! I think it would be much easier.

fastestimator/network.py
```diff
@@ -249,29 +262,35 @@ def forward_step_static(self,
...
+        weights_path: weights path to load from. Defaults None.
     Returns:
         models: model(s) built by FastEstimator
     """
-    models = to_list(model_fn())
-    optimizer_fn = to_list(optimizer_fn)
-    assert len(models) == len(optimizer_fn)
-    for idx, (model, optimizer_def) in enumerate(zip(models, optimizer_fn)):
-        models[idx] = _fe_compile(model, optimizer_def)
+    models, optimizer_fn = to_list(model_fn()), to_list(optimizer_fn),
```
TortoiseHam: extra hanging ','

vbvg2008: how come there's no error before lol

fastestimator/op/tensorop/model/update.py
```diff
@@ -38,6 +38,7 @@ def __init__(self,
                  mode: Union[None, str, Iterable[str]] = "train"):
         super().__init__(inputs=loss_name, outputs=None, mode=mode)
         self.model = model
+        self.loss_name = loss_name
```
TortoiseHam: this should be a setter property to influence the value at inputs[0] rather than a true property. if we modify loss_name somewhere it won't be given the correct input like this

fastestimator/network.py
vbvg2008: done

fastestimator/network.py
vbvg2008: done

fastestimator/backend/load_model.py
vbvg2008: done

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,31 @@
...
+
+    Args:
+        model: model instance
+        save_dir: folder path to save model
+        frequency: model saving frequency in epoch(s). Defaults to 1.
+    """
+    def __init__(self, model: Union[tf.keras.Model, torch.nn.Module], save_dir: str, frequency: int = 1):
+        super().__init__(mode="train")
+        self.model = model
+        self.save_dir = save_dir
```
TortoiseHam: should do the save_dir if check as an assert out here rather than every epoch end

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,31 @@
...
+        self.frequency = frequency
+
+    def on_epoch_end(self, data: Data):
+        if self.save_dir and self.system.epoch_idx % self.frequency == 0:
+            model_name = "epoch_{}".format(self.system.epoch_idx)
```
TortoiseHam: we need more info to go into the model name. what if user training GAN is saving both their generator and discriminator?

fastestimator/trace/io/model_saver.py
```diff
@@ -12,10 +12,31 @@
...
 
 class ModelSaver(Trace):
-    def __init__(self, model):
-        pass
+    """save model weights based on epoch frequency during training
+
+    Args:
+        model: model instance
+        save_dir: folder path to save model
+        frequency: model saving frequency in epoch(s). Defaults to 1.
```
TortoiseHam: users never want to save by step frequency? we could perhaps allow str frequencies like "100s" or "2e"

fastestimator/backend/update_model.py
TortoiseHam: interesting

fastestimator/dataset/data/breast_cancer.py
TortoiseHam: but here we're telling the end user 'this is your training data'. if they want a random split they can just use our .split() syntax later. I would be surprised to find that my training set and eval set were getting mixed up every time a ran an experiment

fastestimator/network.py
vbvg2008: done, now it's much clearer

fastestimator/op/tensorop/model/update.py
vbvg2008: we don't need loss_name here now.

fastestimator/trace/io/model_saver.py
vbvg2008: the purpose of this `save_dir` not for assertion, it is for future integration testing ( when save_dir=None, we won't actually write any model to disk )

TortoiseHam: ah. should probably have an inline comment whenever we're doing something janky for the sake of integration tests so that people don't forget why it's there

fastestimator/trace/io/model_saver.py
vbvg2008: done. now we take model name in fe.build

fastestimator/trace/io/model_saver.py
vbvg2008: I can see the usecase for 100s or 100step.   I guess we will add that in the future if requested, not in this PR tho.

TortoiseHam: k

fastestimator/dataset/data/breast_cancer.py
vbvg2008: I see your point.

fastestimator/trace/io/model_saver.py
vbvg2008: no need for that in my opinion,  it 's pretty obvious

fastestimator/network.py
```diff
@@ -13,11 +13,12 @@
 # limitations under the License.
 # ==============================================================================
 from collections import ChainMap
-from typing import Any, Callable, Dict, Iterable, List, MutableMapping, Set, Union
+from typing import Any, Callable, Dict, Iterable, List, MutableMapping, Optional, Set, Union
```
TortoiseHam: you added the Optional import, but I don't see it being used in any of the relevant places (like List[Optional[str]] for weight path)

fastestimator/trace/io/best_model_saver.py
```diff
@@ -0,0 +1,62 @@
...
+        elif self.save_best_mode == "max":
+            self.best = -np.Inf
+            self.monitor_op = np.greater
+        else:
+            raise ValueError("save_best_mode must be either 'min' or 'max'")
+
+    def on_epoch_end(self, data: Data):
+        if self.save_dir and self.monitor_op(data[self.metric], self.best):
+            self.best = data[self.metric]
+            model_name = "{}_epoch_{}_best_{}".format(self.model.model_name, self.system.epoch_idx, self.metric)
```
TortoiseHam: isn't save best supposed to override outdated saves? should it really be using epoch info in save path?

fastestimator/trace/io/model_saver.py
TortoiseHam: it was not at all obvious to me why we were doing it, esp when the typing for save_dir doesn't indicate that None is even an allowed value

fastestimator/network.py
vbvg2008: done

fastestimator/trace/io/best_model_saver.py
vbvg2008: nice catch

fastestimator/op/tensorop/loss/cross_entropy.py
```diff
@@ -32,18 +32,36 @@ class CrossEntropy(TensorOp):
...
-        loss = cross_entropy(y_pred, y_true, from_logits=self.from_logits, average_loss=self.average_loss)
+        form = self.form
+        if form is None:
+            if len(y_pred.shape) == 2 and y_pred.shape[-1] > 1:
+                if len(y_true.shape) == 2 and y_true.shape[-1] > 1:
+                    form = "categorical"
+                else:
+                    form = "sparse"
+            else:
+                form = "binary"
```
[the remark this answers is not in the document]
geez0219: My fault. It will go to form sparse

fastestimator/trace/io/model_saver.py
vbvg2008: done

PR #486, opened 2020-03-08
update onehot, cifar apphub

vbvg2008:


fastestimator/op/numpyop/univariate/onehot.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Onehot(NumpyOp):
+    """ Transform the label integer to one-hot-encoding
+        Ref: https://towardsdatascience.com/label-smoothing-making-model-robust-to-incorrect-labels-2fae037ffbd0
+
+    Args:
+        num_classes: Total number of classes.
+        label_smoothing: Smoothing factor. The class index value will become: 1 - label_smoothing + label_smoothing
+            / num_classes and The other class index value will become label_smoothing / num_classes
```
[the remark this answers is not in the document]
vbvg2008: nice

fastestimator/backend/get_gradient.py
```diff
@@ -0,0 +1,49 @@
...
+        gradients as sequence of tensors
+    """
+    if tape:
+        with NonContext() if higher_order else tape.stop_recording():
+            gradients = tape.gradient(target, sources)
+    else:
+        gradients = torch.autograd.grad(target,
+                                        sources,
+                                        retain_graph=retain_graph,
+                                        create_graph=higher_order,
```
TortoiseHam: not sure that create_graph = True will result in higher order derivatives. When I run two calls with that flag set the result is the same each time

fastestimator/backend/get_gradient.py
```diff
@@ -0,0 +1,49 @@
...
+# ==============================================================================
+from typing import Iterable, Optional, Union
+
+import tensorflow as tf
+import torch
+
+from fastestimator.util.util import NonContext
+
+
+def get_gradient(target: Union[tf.Tensor, torch.Tensor],
```
TortoiseHam: Lets use the Tensor TypeVar here

fastestimator/backend/get_gradient.py
vbvg2008: it should be the same results, but when create_graph=true, the gradient output will have a .grad field - that's what we are looking for when doing higher order

fastestimator/op/numpyop/univariate/onehot.py
```diff
@@ -0,0 +1,41 @@
...
+        num_classes: Total number of classes.
+        label_smoothing: Smoothing factor, after smoothing class output is: 1 - label_smoothing + label_smoothing
+            / num_classes, the other class output is: label_smoothing / num_classes
+        inputs: Input key(s) of labels to be onehot encoded
+        outputs: Output key(s) of labels
+        mode: What execution mode (train, eval, None) to apply this operation
+    """
+    def __init__(self,
+                 num_classes: int,
+                 label_smoothing: Optional[float] = None,
```
TortoiseHam: instead of making this an optional float why not just default to 0?

fastestimator/op/numpyop/univariate/onehot.py
```diff
@@ -0,0 +1,41 @@
...
+                 outputs: Union[None, str, Iterable[str]] = None,
+                 mode: Union[None, str, Iterable[str]] = None):
+
+        super().__init__(inputs=inputs, outputs=outputs, mode=mode)
+        self.num_classes = num_classes
+        self.label_smoothing = label_smoothing
+        if not self.label_smoothing:
+            self.label_smoothing = 0.0
+
+    def forward(self, data: Union[int, np.ndarray], state: Dict[str, Any]) -> Union[np.ndarray, List[np.ndarray]]:
```
TortoiseHam: this should be written to support multi-io since it's univariate and 'inputs' method signature indicates it accepts Iterable[str] as inputs

fastestimator/op/numpyop/univariate/onehot.py
vbvg2008: good idea, done

fastestimator/backend/get_gradient.py
vbvg2008: done

fastestimator/backend/get_gradient.py
TortoiseHam: when i run an example of this the result tensor .grad is not populated either

fastestimator/op/numpyop/univariate/onehot.py
vbvg2008: multi-IO needs further debate, let's address this later.

PR #499, opened 2020-03-18
enabling negation in modes

vbvg2008:


fastestimator/network.py
```diff
@@ -117,6 +117,12 @@ def _forward_batch(batch: MutableMapping[str, Any], state: Dict[str, Any], ops:
             if op.outputs:
                 write_outputs_by_op(op, batch, data)
 
+    def run_step(self, batch: Dict[str, Any]):
```
TortoiseHam: run step is already defined in the baseNework

fastestimator/network.py
```diff
@@ -117,6 +117,12 @@ def _forward_batch(batch: MutableMapping[str, Any], state: Dict[str, Any], ops:
...
+        raise NotImplementedError
+
+    def transform(self, data: Dict[str, Any], mode: str, epoch: int = 0):
```
TortoiseHam: return type

fastestimator/network.py
vbvg2008: no it's not

fastestimator/network.py
vbvg2008: in baseclass we don't need return type because there's no return. but I added the return type in other classes.

fastestimator/pipeline.py
```diff
@@ -113,6 +118,28 @@ def benchmark(self, mode: str = "train", epoch: int = 0, num_steps: int = 1000,
                 print("FastEstimator: Step: {}, Epoch: {}, Steps/sec: {}".format(idx, epoch, iters_per_sec))
                 start = time.perf_counter()
 
+    def transform(self, data: Dict[str, Any], mode: str, epoch: int = 0):
```
TortoiseHam: what if user puts in batched data? like if they want to inference 4 images at once

fastestimator/network.py
TortoiseHam: oh, I guess I only fixed that in my xai branch. should have return typing:


    def run_step(self, batch: Dict[str, Any]) -> Tuple[Dict[str, Any], Dict[str, Any]]:
        # Batch, Predictions
        raise NotImplementedError

fastestimator/network.py
TortoiseHam: it should be in base so that people know what it's supposed to return when they implement it

fastestimator/pipeline.py
vbvg2008: they will have to call infer 4 times.    the infer is designed to work one data at a time.

fastestimator/network.py
TortoiseHam: also for the sake of IDE which will be looking at the base signature when hinting

fastestimator/op/tensorop/loss/cross_entropy.py
```diff
@@ -38,7 +38,7 @@ class CrossEntropy(TensorOp):
     def __init__(self,
                  inputs: Union[None, str, Iterable[str], Callable] = None,
                  outputs: Union[None, str, Iterable[str]] = None,
-                 mode: Union[None, str, Iterable[str]] = None,
+                 mode: Union[None, str, Iterable[str]] = "!infer",
```
TortoiseHam: we should make UpdateOp mode "!infer" also right?

fastestimator/network.py
vbvg2008: fixed

fastestimator/network.py
vbvg2008: fixed

fastestimator/pipeline.py
TortoiseHam: suppose a user has a model which needs 2 images of horse and 2 images of zebra in order to forward pass

fastestimator/op/tensorop/loss/cross_entropy.py
TortoiseHam: or wait, I guess it's probably only "train" already

fastestimator/pipeline.py
vbvg2008: the 2images of horse and zebra won't be appearing on batch dimension. they will be appearing on different keys like "horse1" "horse2"...

fastestimator/pipeline.py
```diff
@@ -187,6 +222,15 @@ def get_signature_epochs(self, total_epochs: int):
 
     @lru_cache(maxsize=None, typed=True)
     def get_all_output_keys(self, mode: str, total_epochs: int) -> Set[str]:
+        """get the pipeline output keys for a given mode
+
+        Args:
+            mode: current mode, can be "train", "eval" or "test"
```
TortoiseHam: or "infer"

PR #500, opened 2020-03-18
hot fix on effective output keys

vbvg2008:


fastestimator/network.py
```diff
@@ -213,9 +223,9 @@ def transform(self, data: Dict[str, Any], mode: str, epoch: int = 0) -> Dict[str
         """
         self.load_epoch(mode, epoch, warmup=True)
         data = to_tensor(data, "torch")
-        if mode == "infer":
-            self.effective_inputs[mode] = set(data.keys())
-            self.effective_outputs[mode] = set(data.keys())
+        if mode == "infer" or mode not in self.effective_inputs:
```
TortoiseHam: no reason to privilege 'infer' here. just do 'if mode not in self.effective_inputs'

vbvg2008: during "infer", we always want to do that check.    

because the `effective_inputs` and `effective_outputs` may be very different between different epoch of inferencing.

TortoiseHam: then users will face the same problem if they call want to transform(mode=Train) on different epochs and they never created an estimator

PR #506, opened 2020-03-20
Add UNet in apphub

chm123:


fastestimator/op/numpyop/meta/delete.py
```diff
@@ -0,0 +1,32 @@
...
+# limitations under the License.
+# ==============================================================================
+from typing import Any, Dict, Iterable, List, Union
+
+import numpy as np
+
+from fastestimator.op import NumpyOp
+
+
+class Delete(NumpyOp):
```
TortoiseHam: not sure this should be a 'meta' op. This is more like univariate. Meta so far refers to Ops which take other Ops as input

fastestimator/trace/metric/dice.py
```diff
@@ -0,0 +1,68 @@
...
+        true_key: The key of the ground truth mask
+        pred_key: The key of the prediction values
+        mode: What mode to execute in (None to execute in all modes)
+        output_name: What to call the output from this trace (for example in the logger output)
+    """
+    def __init__(self,
+                 true_key: str,
+                 pred_key: str,
+                 threshold: float = 0.5,
+                 mode: Optional[str] = "eval",
```
TortoiseHam: could be list of strings also

TortoiseHam: but also, default should be ("eval", "test") I think

apphub/semantic_segmentation/unet_montgomery/unet_montgomery_tf.py
```diff
@@ -0,0 +1,99 @@
...
+        train_data=csv,
+        eval_data=csv.split(0.2),
+        batch_size=batch_size,
+        ops=[
+            ReadImage(inputs="image", parent_path=csv.parent_path, outputs="image", grey_scale=True),
+            ReadImage(inputs="mask_left",
+                      parent_path=csv.parent_path,
+                      outputs="mask_left",
+                      grey_scale=True,
+                      mode='!infer'),
```
TortoiseHam: why would you not read image during inference?

chm123: During inferencing, I don't have masks. Only image is sent to the network.

fastestimator/op/numpyop/meta/delete.py
vbvg2008: we should extend this definition of meta.    I also have BatchOp that I put in meta folder.

chm123: Or create a new category?

fastestimator/trace/metric/dice.py
chm123: fixed

apphub/semantic_segmentation/unet_montgomery/unet_montgomery_tf.py
TortoiseHam: oh right

fastestimator/op/numpyop/meta/delete.py
TortoiseHam: I think those are both examples of univariate. they take one or more keys and act on them each individually

PR #507, opened 2020-03-20
Added siamese network for tensoflow to apphub

gaurav272333:
Added tf version of siamese network for apphub


apphub/one_shot_learning/siamese_network_tf.py
```diff
@@ -0,0 +1,192 @@
...
+        self.trials = trials
+
+    def on_epoch_begin(self, data: Data):
+        self.total = 0
+        self.correct = 0
+
+    def on_epoch_end(self, data: Data):
+        for _ in range(self.trials):
+            img_path = self.dataset.one_shot_trial(self.N)
+            input_img = (np.array([np.expand_dims(cv2.imread(i, cv2.IMREAD_GRAYSCALE), -1) / 255. for i in img_path[0]],
```
TortoiseHam: instead of redefining all the the pre-processing stuff and then calling feed_forward, why not use self.system.network.transform()?

TortoiseHam: oh wait, I guess what you really need would be pipeline. should we add that to system?

PR #508, opened 2020-03-20
adding BatchDataset and re-organizing our imports

vbvg2008:


fastestimator/dataset/data/horse2zebra.py
```diff
@@ -62,24 +61,21 @@ def load_data(root_dir: Optional[str] = None) -> Tuple[UnpairedDataset, Unpaired
...
+    train_b = DirDataset(root_dir=os.path.join(data_folder_path, 'trainB'),
+                         data_key="B",
+                         file_extension='.jpg',
+                         recursive_search=False)
 
-    train = UnpairedDataset(train_a, train_b)
-    test = UnpairedDataset(test_a, test_b)
-
-    return train, test
+    return train_a, train_b, test_a, test_b
```
TortoiseHam: Make these into BatchDatasets

fastestimator/dataset/data/horse2zebra.py
```diff
@@ -19,14 +19,13 @@
...
-from fastestimator.dataset.unlabeled_dir_dataset import UnlabeledDirDataset
-from fastestimator.dataset.unpaired_dataset import UnpairedDataset
+from fastestimator.dataset.dir_dataset import DirDataset
 from fastestimator.util.wget_util import bar_custom, callback_progress
 
 wget.callback_progress = callback_progress
 
 
-def load_data(root_dir: Optional[str] = None) -> Tuple[UnpairedDataset, UnpairedDataset]:
+def load_data(root_dir: Optional[str] = None) -> Tuple[DirDataset]:
```
TortoiseHam: -> Tuple[BatchDataset, BatchDataset]

(tuple typing should reflect the exact number of args in the tuple)

fastestimator/dataset/op_dataset.py
```diff
@@ -15,29 +15,39 @@
 from copy import deepcopy
 from typing import List
 
+import numpy as np
 from torch.utils.data import Dataset
 
-from fastestimator.dataset.unpaired_dataset import UnpairedDataset
-from fastestimator.op import NumpyOp, get_inputs_by_op, write_outputs_by_op
+from fastestimator.dataset.batch_dataset import BatchDataset
+from fastestimator.op.op import NumpyOp, get_inputs_by_op, write_outputs_by_op
```
TortoiseHam: why are we using fe.op.op for NumpyOp but using fe.util rather than fe.util.util?

fastestimator/op/numpyop/univariate/__init__.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+# see a complete list of effect in albumentation page: https://albumentations-demo.herokuapp.com
```
TortoiseHam: effects plural. also, this notice should probably go in NumpyOp since the page also includes multi-variate augmentations ('professional mode')

fastestimator/op/op.py
```diff
@@ -18,7 +18,7 @@
 import tensorflow as tf
 import torch
 
-from fastestimator.schedule import Scheduler
+from fastestimator.schedule.schedule import Scheduler
```
TortoiseHam: schedule file is at same level as the schedule __init__, so we should import directly from fe.schedule

fastestimator/op/tensorop/loss/cross_entropy.py
```diff
@@ -17,7 +17,9 @@
 import tensorflow as tf
 import torch
 
-from fastestimator.backend import binary_crossentropy, categorical_crossentropy, sparse_categorical_crossentropy
+from fastestimator.backend.binary_crossentropy import binary_crossentropy
```
TortoiseHam: all of these backend functions are at same level as the backend __init__ file, so they should import directly form fe.backend

fastestimator/pipeline.py
```diff
@@ -62,11 +63,12 @@ def _verify_inputs(self, **kwargs):
     def _verify_dataset(self, mode: str, dataset: DataSource, **kwargs) -> bool:
         if isinstance(dataset, Dataset):
             # batch_size check
-            assert isinstance(self.batch_size, (Scheduler, int)), \
+            assert isinstance(self.batch_size, (Scheduler, int)) or self.batch_size is None, \
```
TortoiseHam: assert isinstance(self.batch_size, (Scheduler, int, type(None)))

saves on the extra or check

fastestimator/pipeline.py
```diff
@@ -62,11 +63,12 @@ def _verify_inputs(self, **kwargs):
...
                 "unsupported batch_size format: {}".format(self.batch_size)
             if isinstance(self.batch_size, Scheduler):
                 for batch_size in self.batch_size.get_all_values():
-                    assert isinstance(batch_size, int), "unsupported batch_size format: {}".format(self.batch_size)
+                    assert isinstance(batch_size, int) or batch_size is None, \
```
TortoiseHam: assert isinstance(self.batch_size, (Scheduler, int, type(None)))

fastestimator/pipeline.py
```diff
@@ -184,11 +186,14 @@ def get_loader(self, mode: str, epoch: int = 0,
             batch_size = self.batch_size
             if isinstance(batch_size, Scheduler):
                 batch_size = batch_size.get_current_value(epoch)
+            if isinstance(data, BatchDataset) and not batch_size is None:
```
TortoiseHam: and batch_size is not None

more readable

fastestimator/trace/adapt/lr_scheduler.py
```diff
@@ -19,8 +19,9 @@
 import tensorflow as tf
 import torch
 
-from fastestimator.backend import get_lr, set_lr
-from fastestimator.summary import System
+from fastestimator.backend.get_lr import get_lr
```
TortoiseHam: again, user should never be drilling into the actual file to get the imports

fastestimator/trace/metric/confusion_matrix.py
```diff
@@ -20,7 +20,7 @@
 
 from fastestimator.backend.to_number import to_number
 from fastestimator.trace.trace import Trace
-from fastestimator.util import Data
+from fastestimator.util.data import Data
```
TortoiseHam: inconsistent with fact that Data was put into util init file

fastestimator/pipeline.py
```diff
@@ -184,11 +186,14 @@ def get_loader(self, mode: str, epoch: int = 0,
...
+                raise ValueError("batch_size must be None when using BatchDataset")
```
TortoiseHam: don't we want to just ignore the batch size and raise a warning? like if they use a scheduler and have non-batch dataset epoch 0-5 and then batch dataset after that

fastestimator/dataset/batch_dataset.py
```diff
@@ -0,0 +1,132 @@
...
+        # Overwriting the split() method instead of _do_split
+        raise AssertionError("This method should not have been invoked. Please file a bug report")
+
+    def split(self, *fractions: Union[float, int, Iterable[int]]) -> Union['UnpairedDataset', List['UnpairedDataset']]:
+        new_datasets = [to_list(ds.split(*fractions)) for ds in self.datasets]
+        num_splits = len(new_datasets[0])
+        new_datasets = [[ds[i] for ds in new_datasets] for i in range(num_splits)]
+        results = [BatchDataset(ds, self.num_samples, self.probability) for ds in new_datasets]
+        # Re-compute personal variables
+        self.index_maps = [list(range(len(dataset))) for dataset in self.datasets]
```
TortoiseHam: use max(len(datasets)) for each and later put the mod on the outside

fastestimator/pipeline.py
vbvg2008: this batch_size is the batch_size on that specific epoch, it won't be a scheduler.

TortoiseHam: so what happens right now if I pass batch_size = 20 but use a scheduler that contains both a tf.dataset and an fe.dataset?

fastestimator/dataset/data/horse2zebra.py
vbvg2008: batch_size is required to use `BatchDatasets` in this case. It makes more sense to initialize `BatchDataset` under the `get_estimator` where user can leverage the `batch_size` argument.

fastestimator/dataset/data/horse2zebra.py
vbvg2008: fixed

fastestimator/dataset/op_dataset.py
vbvg2008: changed everything to folder import

fastestimator/dataset/data/horse2zebra.py
TortoiseHam: in that case lets make batch_size an input arg of the horse2zebra function

fastestimator/op/numpyop/univariate/__init__.py
vbvg2008: added this notice in multivariate init as well.

fastestimator/dataset/data/horse2zebra.py
vbvg2008: done

fastestimator/op/op.py
vbvg2008: yep,done

fastestimator/op/tensorop/loss/cross_entropy.py
vbvg2008: done

fastestimator/pipeline.py
vbvg2008: nice, done

fastestimator/pipeline.py
vbvg2008: nice done

fastestimator/pipeline.py

fastestimator/trace/adapt/lr_scheduler.py
vbvg2008: done

fastestimator/trace/metric/confusion_matrix.py
vbvg2008: done

fastestimator/pipeline.py
vbvg2008: we will raise a warning saying batch_size will only be used for built-in dataset

fastestimator/dataset/batch_dataset.py
vbvg2008: like this?

fastestimator/dataset/batch_dataset.py
```diff
@@ -0,0 +1,132 @@
...
+    def __init__(self,
+                 datasets: Union[FEDataset, Iterable[FEDataset]],
+                 num_samples: Union[int, Iterable[int]],
+                 probability: Optional[Iterable[float]] = None):
+        self.datasets = to_list(datasets)
+        self.num_samples = to_list(num_samples)
+        self.probability = to_list(probability)
+        self.same_feature = False
+        self._check_input()
+        self.index_maps = [list(range(max((len(dataset) for dataset in self.datasets)))) for _ in self.datasets]
```
TortoiseHam: technically you might want to compute the max first to reduce looping, but realistically this won't matter

fastestimator/dataset/data/horse2zebra.py
```diff
@@ -19,23 +19,23 @@
...
-from fastestimator.dataset.unpaired_dataset import UnpairedDataset
-from fastestimator.util.wget_util import bar_custom, callback_progress
+from fastestimator.dataset import BatchDataset, DirDataset
+from fastestimator.util import bar_custom, callback_progress
 
 wget.callback_progress = callback_progress
 
 
-def load_data(root_dir: Optional[str] = None) -> Tuple[UnpairedDataset, UnpairedDataset]:
+def load_data(batch_size, root_dir: Optional[str] = None) -> Tuple[BatchDataset, BatchDataset]:
```
TortoiseHam: batch_size: int

vbvg2008: done

PR #513, opened 2020-03-24
 some more changes

vbvg2008:
* use deep copy in `pipeline transfer`

* initialize  step, epoch, batch_idx from 1

* enable `None` as optimizer_fn


fastestimator/backend/feed_forward.py
```diff
@@ -15,17 +15,22 @@
...
 import torch
 
+from fastestimator.backend.to_tensor import to_tensor
+
 Tensor = TypeVar('Tensor', tf.Tensor, torch.Tensor)
 
 
 def feed_forward(model: Union[tf.keras.Model, torch.nn.Module], x: Tensor, training: bool = True) -> Tensor:
     if isinstance(model, tf.keras.Model):
+        if not isinstance(x, tf.Tensor):
```
TortoiseHam: why are we doing this here? will add overhead to all training

fastestimator/pipeline.py
```diff
@@ -112,43 +114,47 @@ def benchmark(self, mode: str = "train", epoch: int = 0, num_steps: int = 1000,
             loader = loader.take(num_steps)
         start = time.perf_counter()
         for idx, _ in enumerate(loader):
-            if idx == num_steps:
-                break
-            if idx % log_interval == 0 and idx > 0:
+            if idx + 1 % log_interval == 0:
```
TortoiseHam: enumerate takes a second arg saying where to start the idx. enumerate(loader, 1) rather than adjusting the idx after the fact

fastestimator/pipeline.py
```diff
@@ -112,43 +114,47 @@ def benchmark(self, mode: str = "train", epoch: int = 0, num_steps: int = 1000,
...
             data: Input data in dictionary format
             mode: Current mode, can be "train", "eval", "test" or "infer"
-            epoch: Current epoch index. Defaults to 0.
+            epoch: Current epoch index. Defaults to 1.
 
         Returns:
             transformed data
         """
+        data = deepcopy(data)
         ops = get_current_ops(self.ops, mode, epoch)
```
TortoiseHam: we should abstract this into a single function somewhere so that pipeline and op dataset don't accidentally get out of sync again

fastestimator/summary/system.py
```diff
@@ -43,31 +43,40 @@ def __init__(self,
 
         self.network = network
         self.mode = mode
-        self.global_step = 0
+        self.global_step = None
```
TortoiseHam: what is the point of None here? if not absolutely necessary better to avoid, and if you do change, need to change the typing info above to Optional[int]

fastestimator/backend/feed_forward.py
TortoiseHam: moreover, if we're allowing user to pass something that isn't a Tensor as 'x', we need to update the typing

TortoiseHam: (but again, we shouldn't do that)

fastestimator/summary/system.py
vbvg2008: with None, it's easier to tell whether the global step is initialized or not.    fixed the typing.

fastestimator/pipeline.py
vbvg2008: `pipeline.transform` and `network._forward_batch` and `opdataset._forward` are all very different. doesn't make much sense to wrap them.

fastestimator/backend/feed_forward.py
vbvg2008: it won't affect the training speed from my benchmark.    My intent is to enable `feed_forward` for model prediction.

vbvg2008: fixed the typing

fastestimator/pipeline.py
vbvg2008: fixed

fastestimator/pipeline.py

fastestimator/schedule/schedule.py
```diff
@@ -44,10 +44,10 @@ def __init__(self, epoch_dict: Dict[int, T]):
         assert isinstance(epoch_dict, dict), "must provide dictionary as epoch_dict"
         self.epoch_dict = epoch_dict
         self.keys = sorted(self.epoch_dict)
-        assert 0 in self.epoch_dict, "epoch 0 is missing in dictionary, use None if no op is needed"
+        assert 1 in self.epoch_dict, "epoch 1 is missing in dictionary, use None if no op is needed"
         for key in self.keys:
             assert isinstance(key, int), "found non-integer key: {}".format(key)
-            assert key >= 0, "found negative key: {}".format(key)
+            assert key > 1, "found non-positive key: {}".format(key)
```
TortoiseHam: key >=1 right?

fastestimator/trace/trace.py
```diff
@@ -90,9 +90,15 @@ def on_epoch_begin(self, data: Data):
             self.epoch_start = time.perf_counter()
             self.step_start = time.perf_counter()
 
+    def on_batch_begin(self, data: Data):
```
TortoiseHam: thought we were going to move this back?

fastestimator/schedule/schedule.py
vbvg2008: fixed

fastestimator/trace/trace.py
vbvg2008: just moved it

PR #526, opened 2020-03-28
new nightly_test template: add need_test argument

geez0219:
Old nightly test python file, will only call < fastestimator train ....>
New version will also call <fastestimator test ...> if need_test = True


geez0219:
New update: Fix the jenkins tag  error in the README.md file. The Jenkins nightly-build tag is linked to new 1.0dev pipeline

PR #528, opened 2020-03-31
Added nightly test script and modified apphub files

gaurav272333:
Following changes done for Siamese apphub
1. Added parameters tag to jupyter notebook
2. Added max_steps_per_epoch parameter to tf and torch scripts
3. Added nightly test scripts


geez0219:
Need to move the save_dir variable in the parameters cell in order to follow the convention
(I will fix this after merge).

PR #529, opened 2020-03-31
Added numpyop to read .mat files

gaurav272333:
NumpyOp for reading .mat files.


vbvg2008:
Plz use the new matreader

PR #531, opened 2020-03-31
new mat reader

vbvg2008:


fastestimator/op/numpyop/multivariate/load_mat.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+from typing import Any, Dict, Iterable, List, Union
+
+from scipy.io import loadmat
+
+from fastestimator.op.numpyop.numpyop import NumpyOp
+from fastestimator.util.util import to_list
+
+
+class LoadMat(NumpyOp):
```
TortoiseHam: should we name this ReadMat to have consistency with ReadImage?

fastestimator/op/numpyop/multivariate/load_mat.py
```diff
@@ -0,0 +1,49 @@
...
+        file_in: Key of file path to be loaded.
+        keys_in: Key(s) of dictionaries to read.
+        keys_out: Keys(s) of dictionaries to output.
+        mode: What execution mode (train, eval, test, None) to apply this operation
+        parent_path: Parent path that will be prepended to a given filepath
+    """
+    def __init__(self,
+                 file_in: str,
+                 keys_in: Union[str, Iterable[str]],
+                 keys_out: Union[str, Iterable[str]],
```
TortoiseHam: we could make this univariate by converting file_in -> inputs, keys_out -> outputs, and removing keys_in from the function signature

PR #536, opened 2020-04-06
Added Multi task Learning to apphub

gaurav272333:
Added multi task learning to apphub and nightly test scripts


apphub/multi_task_learning/uncertainty_weighted_loss/uncertainty_loss_tf.py
```diff
@@ -0,0 +1,243 @@
...
+from fastestimator.trace.metric import Accuracy, Dice
+
+
+class ReduceLoss(TensorOp):
+    """TensorOp to average loss for a batch"""
+    def forward(self, data, state):
+        return reduce_loss(data)
+
+
+class UncertaintyWeightedLoss(layers.Layer):
```
TortoiseHam, line 44: is this not generic enough to make into a re-usable layer?

vbvg2008: it is just one idea from one paper, i don't think the usage of it is general enough.

PR #537, opened 2020-04-06
make nightly-test download dataset from s3

geez0219:


geez0219:
update the new commit that use sync for copying

PR #637, opened 2020-05-08
remove dollar sign and add plot section in docstring of Trace

geez0219:
1. cuurently our website pluging cannot support render the $ in the markdown. (it will get confused with Katex mathmetical expression). Therefore before we finish working on that plugin issue, we need to temporarily don't use $

2. For the illustration of Trace docstring, we need to put a ```plot ``` coding block around it.


TortoiseHam:
This PR also needs to be opened vs the R1.0 branch

PR #655, opened 2020-05-12
fix batch dataset

vbvg2008:


TortoiseHam:
Do you want that Network fix in this PR also?

fastestimator/dataset/batch_dataset.py
```diff
@@ -63,7 +63,7 @@ def __init__(self,
         self.probability = to_list(probability)
         self.same_feature = False
         self._check_input()
-        self.index_maps = [list(range(max((len(dataset) for dataset in self.datasets)))) for _ in self.datasets]
+        self.index_maps = [list(range(sum((len(dataset) for dataset in self.datasets)))) for _ in self.datasets]
```
TortoiseHam, line 66: What is this intended to solve? we're already taking the index values mod the length of the datasets

PR #685, opened 2020-06-01
backend unit test

Vivek305:


geez0219:
We had agreement not to add this:

if __name__ == "__main__":
    unittest.main()

PR #701, opened 2020-06-05
added lambda op and test for op.py

vbvg2008:


fastestimator/op/op.py
```diff
@@ -104,3 +104,18 @@ def write_outputs_by_op(op: Op, store: MutableMapping[str, Any], outputs: Any) -
         outputs = [outputs]
     for key, data in zip(op.outputs, outputs):
         store[key] = data
+
+
+@traceable()
+class LambdaOp(Op):
+    def __init__(self,
```
TortoiseHam, line 111: doc string

vbvg2008: done

PR #707, opened 2020-06-08
remove lambda function as inputs, use lambdaOp instead

vbvg2008:


fastestimator/op/op.py
```diff
@@ -128,4 +128,4 @@ def __init__(self,
         self.in_list = True
 
     def forward(self, data: List[Any], state: Dict[str, Any]) -> Any:
-        return self.fn(*data)
+        return self.fn(*to_list(data))
```
TortoiseHam, line 131: no need to call to_list when in_list = True

vbvg2008: When inputs is None, the data becomes None.

TortoiseHam: that's a problem - should be resolved elsewhere such that in_list = True always results in a list being passed

vbvg2008: like this?

fastestimator/pipeline.py
```diff
@@ -21,15 +21,15 @@
...
-from torch.utils.data.dataloader import default_collate
-
 from fastestimator.dataset.batch_dataset import BatchDataset
 from fastestimator.dataset.op_dataset import OpDataset
 from fastestimator.op.numpyop.numpyop import NumpyOp, forward_numpyop
+from fastestimator.op.op import LambdaOp
 from fastestimator.schedule.schedule import Scheduler, get_current_items
 from fastestimator.util.traceability_util import traceable
 from fastestimator.util.util import pad_batch, to_list, to_set
+from torch.utils.data import DataLoader, Dataset, RandomSampler
```
TortoiseHam, line 31: import sort misconfigured

PR #733, opened 2020-06-16
Model Summary Traceability

TortoiseHam:
remove the early stop (+5 squashed commits)
Squashed commits:
[2073738] make tables look better
[73ae8f5] consolidate report resources into a folder so that it can be rebuilt later
[c729607] TF and Torch model summaries
[e46cdd6] first pass at PyTorch model traceability
[2b94fbb] tensorflow models

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,9 +50,13 @@ def get_dependency():
         'jsonpickle',
         'python-docx',
         'PyLaTeX',
+        'natsort',
         'tensorflow_probability==0.8.0',
         'transformers==2.4.1',
-        'tensorflow==2.1.0'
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -50,9 +50,13 @@ def get_dependency():
         'jsonpickle',
         'python-docx',
         'PyLaTeX',
+        'natsort',
         'tensorflow_probability==0.8.0',
         'transformers==2.4.1',
-        'tensorflow==2.1.0'
+        'tensorflow==2.1.0',
+        'pytorch-model-summary',
```
vbvg2008, line 57: need specific version for these libraries.

PR #743, opened 2020-06-21
Finished RetinaNet application hub

vbvg2008:
* tensorflow implementation produced 0.264 mAP on image size 512 (with default batch 8)
* pytorch implementation produced 0.271 mAP on image size 512 (with default batch 16)
* major refactor in both implementation
* added retinanet in nightly test


chm123:
LGTM

PR #763, opened 2020-06-29
finish unittest of estimator, qms trace (passed all test in window os)

geez0219:
1. finish unit test of estimator, qms trace, sort_trace function in trace.py
2. modify test_pipeline
3. upgrade the is_equal function with the new feature allowing users to check dtype
4. add some new test for is_equal function
5. add __init__.py files to all other missing directories


fastestimator/estimator.py
```diff
@@ -233,6 +233,7 @@ def _start(self, run_modes: Set[str]) -> None:
             run_modes: The current execution modes.
         """
         all_traces = sort_traces(get_current_items(self.traces_in_use, run_modes=run_modes))
+
```
vbvg2008, line 236: remove this extra line

PR #777, opened 2020-07-08
extend our cosine decay to allow restart

vbvg2008:


fastestimator/schedule/lr_shedule.py
```diff
@@ -24,19 +29,33 @@ def cosine_decay(time: int, cycle_length: int, init_lr: float, min_lr: float = 1
...
 
+    For more information, checkout SGDR: https://arxiv.org/pdf/1608.03983.pdf.
+
     Args:
         time: The current step or epoch during training starting from 1.
         cycle_length: The decay cycle length.
         init_lr: Initial learning rate to decay from.
         min_lr: Minimum learning rate.
         start: The step or epoch to start the decay schedule.
+        cycle_multiplier: The factor of which next cycle length will be multiplied.
```
TortoiseHam, line 40: The factor by which the next cycle length will be multiplied.

fastestimator/schedule/lr_shedule.py
```diff
@@ -24,19 +29,33 @@ def cosine_decay(time: int, cycle_length: int, init_lr: float, min_lr: float = 1
     fe.Estimator(..., traces=[s])
     ```
 
+    For more information, checkout SGDR: https://arxiv.org/pdf/1608.03983.pdf.
```
TortoiseHam, line 32: check out

fastestimator/schedule/lr_shedule.py
```diff
@@ -24,19 +29,33 @@ def cosine_decay(time: int, cycle_length: int, init_lr: float, min_lr: float = 1
...
+        time = time - start + 1
+        if cycle_multiplier > 1:
+            current_cycle_idx = math.ceil(
+                math.log(time * (cycle_multiplier - 1) / cycle_length + 1) / math.log(cycle_multiplier)) - 1
+            cumulative = cycle_length * (cycle_multiplier**current_cycle_idx - 1) / (cycle_multiplier - 1)
+        elif cycle_multiplier == 1:
+            current_cycle_idx = math.ceil(time / cycle_length) - 1
+            cumulative = current_cycle_idx * cycle_length
+        else:
+            raise ValueError("multiplier must be at least 1")
```
TortoiseHam, line 57: why do we enforce that the user can't shorten the cycles?

vbvg2008: it doesn't make sense to do that.   we want lr to become more and more stable as training progresses

fastestimator/schedule/lr_shedule.py
vbvg2008: y

fastestimator/schedule/lr_shedule.py
vbvg2008: y

PR #806, opened 2020-07-20
Added suggested modifications in Op benchmark

gaurav272333:


fastestimator/pipeline.py
```diff
@@ -179,7 +179,7 @@ def benchmark(self, mode: str = "train", epoch: int = 1, num_steps: int = 1000,
             if self.batch_size:
                 log_interval = log_interval * self.batch_size
 
-            print("\nBreakdown of time taken by Pipeline Operations:")
+            print("\nBreakdown of time taken by Pipeline Operations: Epoch: {}, Mode: {}".format(epoch, mode))
```
TortoiseHam, line 182: Pipeline Operations (train epoch 7):

rather than

Pipeline Operations: Epoch 7: Mode: train

fastestimator/pipeline.py
```diff
@@ -201,9 +201,19 @@ def benchmark(self, mode: str = "train", epoch: int = 1, num_steps: int = 1000,
                         duration_list[i] += duration
 
             total_time = np.sum(duration_list)
+            max_op_len = max(len(op.__class__.__name__) for op in op_list)
+            max_in_len = max([len(str(op.inputs)) - 2 for op in op_list] + [len("Inputs")])
```
TortoiseHam, line 205: could save some print characters by doing ", ".join(op.inputs) rather than str(op.inputs)

gaurav272333: Sure. I will make that change.

fastestimator/pipeline.py
gaurav272333: Okay. I will update that.

PR #815, opened 2020-07-24
add max_to_keep argument to ModelSaver

geez0219:
1. add a new argument to ModelSaver: max_to_keep.
    It allows users to set an upper bound of file number to keep that generated from the trace

2. add new unit test for this feature

### DIFF
--- a/fastestimator/trace/io/model_saver.py
+++ b/fastestimator/trace/io/model_saver.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
-from typing import Union
+import os
+import queue
+from typing import Optional, Union
 
 import tensorflow as tf
 import torch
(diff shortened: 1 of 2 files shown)


fastestimator/trace/io/model_saver.py
```diff
@@ -31,16 +33,32 @@ class ModelSaver(Trace):
...
+    def __init__(self,
+                 model: Union[tf.keras.Model, torch.nn.Module],
+                 save_dir: str,
+                 frequency: int = 1,
+                 max_to_keep: Optional[int] = None) -> None:
         super().__init__(mode="train")
         self.model = model
         self.save_dir = save_dir
         self.frequency = frequency
+        self.max_to_keep = max_to_keep
```
TortoiseHam, line 47: queue.Queue() is a lot of overhead for this, and a lot of extra variables. instead eliminate self.max_to_keep and the 'if max_to_keep' condition here and...

from collections import deque
...
self.file_queue = deque([None] * (max_to_keep or 0), maxlen = max_to_keep or 0)

fastestimator/trace/io/model_saver.py
```diff
@@ -31,16 +33,32 @@ class ModelSaver(Trace):
...
+            self.file_queue = queue.Queue()
 
     def on_epoch_end(self, data: Data) -> None:
         # No model will be saved when save_dir is None, which makes smoke test easier.
         if self.save_dir and self.system.epoch_idx % self.frequency == 0:
             model_name = "{}_epoch_{}".format(self.model.model_name, self.system.epoch_idx)
             model_path = save_model(self.model, self.save_dir, model_name)
             print("FastEstimator-ModelSaver: Saved model to {}".format(model_path))
+
+            if self.max_to_keep:
```
TortoiseHam, line 58: replace all of this with:
`
rm_path = self.file_queue[self.file_queue.maxlen -1] if self.file_queue.maxlen else None
if rm_path:
  os.remove(os.path.join(self.save_dir, rm_path))
self.file_queue.appendleft(model_path)
`

TortoiseHam: as a side note, do you need to join the save_dir with the path? doesn't save model give the full path already? if not, we probably should change the save model backend function to do that

geez0219: You are right, the os.path.join(self.save_dir, rm_path) is totally unnecessary. I will drop this.

geez0219:
Finished the modification following Michael's comments. The updated code is not 100% the same as the comments because  
code in comments and previous code don't handle the max_to_keep = 0 correctly (new code required it to be larger than 0 )

PR #816, opened 2020-07-24
Added trace TerminateOnNaN

gaurav272333:


vbvg2008:
remember to add unit tests for them

fastestimator/trace/adapt/terminate_on_nan.py
```diff
@@ -0,0 +1,90 @@
...
+from fastestimator.schedule.schedule import get_current_items
+from fastestimator.trace.trace import Trace
+from fastestimator.util.data import Data
+from fastestimator.util.traceability_util import traceable
+from fastestimator.util.util import to_set
+
+
+@traceable()
+class TerminateOnNaN(Trace):
+    """End Training if a NaN value is detected. By default (inputs=None) it will monitor all loss values at the end
```
TortoiseHam, line 30: First Line of doc string is brief summary (can't be more than 1 line). Extra details go in subsequent paragraph

fastestimator/trace/adapt/terminate_on_nan.py
```diff
@@ -0,0 +1,90 @@
...
+    and/or the keys corresponding to the outputs of other traces (ex. accuracy).
+
+    Args:
+        monitor_names: key(s) to monitor for NaN values. If None, all loss values will be monitored. "*" will monitor
+            all trace output keys and losses.
+        mode: What mode(s) to execute this Trace in. For example, "train", "eval", "test", or "infer". To execute
+            regardless of mode, pass None. To execute in all modes except for a particular one, you can pass an argument
+            like "!infer" or "!train".
+    """
+    def __init__(self, monitor_names: Union[None, str, Set[str]] = None, mode: Union[None, str,
```
TortoiseHam, line 41: monitor_names typing should be Iterable[str] rather than Set[str] probably

fastestimator/trace/adapt/terminate_on_nan.py
```diff
@@ -0,0 +1,90 @@
...
+    def on_epoch_end(self, data: Data) -> None:
+        for key in self.monitored_trace_keys:
+            if key in data:
+                if self._check_nan(data[key]):
+                    self.system.stop_training = True
+                    print("FastEstimator-TerminateOnNaN: NaN Detected in: {}".format(key))
+
+    @staticmethod
+    def _check_nan(val: Union[int, float, np.ndarray, tf.Tensor, torch.Tensor]) -> bool:
+        if isinstance(val, tf.Tensor):
```
TortoiseHam, line 85: check_nan should be a backend function rather than a member variable of the trace

fastestimator/trace/adapt/terminate_on_nan.py
```diff
@@ -0,0 +1,90 @@
...
+                    self.system.stop_training = True
+                    print("FastEstimator-TerminateOnNaN: NaN Detected in Loss: {}".format(key))
+        for key in self.monitored_trace_keys:
+            if key in data:
+                if self._check_nan(data[key]):
+                    self.system.stop_training = True
+                    print("FastEstimator-TerminateOnNaN: NaN Detected in: {}".format(key))
+
+    def on_epoch_end(self, data: Data) -> None:
+        for key in self.monitored_trace_keys:
```
TortoiseHam, line 77: There might be NaN in evaluation loss. why only trace_keys here?

fastestimator/trace/adapt/terminate_on_nan.py
```diff
@@ -0,0 +1,90 @@
...
+        mode: What mode(s) to execute this Trace in. For example, "train", "eval", "test", or "infer". To execute
+            regardless of mode, pass None. To execute in all modes except for a particular one, you can pass an argument
+            like "!infer" or "!train".
+    """
+    def __init__(self, monitor_names: Union[None, str, Set[str]] = None, mode: Union[None, str,
+                                                                                     Set[str]] = None) -> None:
+        self.monitored_keys = monitor_names if monitor_names is None else to_set(monitor_names)
+        super().__init__(inputs=self.monitored_keys, mode=mode)
+        self.all_loss_keys = {}
+        self.monitored_loss_keys = {}
```
TortoiseHam, line 46: self.loss_keys, self.trace_keys can be combined into self.monitor_keys which gets built every epoch. self.monitored_keys can be removed since it is encoded inside self.inputs

fastestimator/trace/adapt/terminate_on_nan.py
gaurav272333: I will update the doc string. Thanks.

fastestimator/trace/adapt/terminate_on_nan.py
gaurav272333: Makes Sense. I'll move it to backend.

fastestimator/trace/adapt/terminate_on_nan.py
gaurav272333: The ideas was that on_batch_end method will handle nan losses in eval mode (similar to train data) but some traces outputs could only be available on epoch_end, so checking for trace keys on_epoch_end. Hence, the case. However, we can check that at epoch end too and in that case I can combine both losses and trace keys as well.

fastestimator/trace/adapt/terminate_on_nan.py
gaurav272333: Sure. I will make this change.

fastestimator/trace/adapt/terminate_on_nan.py
gaurav272333: Iterable[str] is better. I will update that. Thanks.

vbvg2008:
still need tests

fastestimator/trace/adapt/terminate_on_nan.py
```diff
@@ -0,0 +1,67 @@
...
+from fastestimator.trace.trace import Trace
+from fastestimator.util.data import Data
+from fastestimator.util.traceability_util import traceable
+
+
+@traceable()
+class TerminateOnNaN(Trace):
+    """End Training if a NaN value is detected.
+
+    By default (inputs=None) it will monitor all loss values at the end of each batch. If one or more inputs are
```
TortoiseHam, line 28: By default (monitor_names=None)

PR #819, opened 2020-07-24
cutmix batch op

Vivek305:
- backend functions- roll and random mix patch
- cutmix batch


fastestimator/backend/random_mix_patch.py
```diff
@@ -0,0 +1,61 @@
...
+
+Tensor = TypeVar('Tensor', tf.Tensor, torch.Tensor)
+
+
+def random_mix_patch(tensor: Tensor, lam: Tensor, uniform_sample: Tensor) -> Tensor:
+    """Randomly cut the patches from input images.
+
+    Args:
+        tensor: The input value.
+        lam: Generated mixed sample.
```
TortoiseHam, line 28: scribe what these values are used for. if this function could be invoked using a number not drawn from a  uniform distribution, then the arg name shouldn't be 'uniform_sample'

also, doc string lines should all end with '.'

fastestimator/backend/random_mix_patch.py
```diff
@@ -0,0 +1,61 @@
...
+        uniform_sample: Sample drawn from the uniform distribution
+
+    Returns:
+        The X and Y coordinates of the patch along with width and height
+
+    Raises:
+        ValueError: If `tensor` is an unacceptable data type.
+    """
+    if tf.is_tensor(tensor):
```
TortoiseHam, line 37: I think this class is kind of missing the point of our backend folder. it ought to be possible to implement this logic in Op without using if branching by using backend calls for maximum and round

fastestimator/backend/roll.py
```diff
@@ -0,0 +1,76 @@
...
+    b = fe.backend.roll(p, shift=1, axis=0)  # [[5, 6, 7], [1, 2, 3]]
+    b = fe.backend.roll(p, shift=2, axis=1)  # [[2, 3, 1], [6, 7, 5]]
+    b = fe.backend.roll(p, shift=-2, axis=1)  # [[3, 1, 2], [7, 5, 6]]
+    b = fe.backend.roll(p, shift=[-1, -1], axis=[0, 1])  # [[6, 7, 5], [2, 3, 1]]
+    ```
+
+    Args:
+        tensor: The input value.
+        shift: The number of places by which the elements need to be shifted. If shift is a list, axis must be a list of
+            same size
```
TortoiseHam, line 60: doc strings end with '.'

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,77 @@
...
+import torch
+
+from fastestimator.backend import random_mix_patch, roll
+from fastestimator.op.tensorop import TensorOp
+
+Tensor = TypeVar('Tensor', tf.Tensor, torch.Tensor)
+
+
+class CutMixBatch(TensorOp):
+    """This class perform cutmix augmentation on batch of tensors.
```
TortoiseHam, line 14: performs

TortoiseHam: on a batch

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,77 @@
...
+        inputs: key of the input to be cut-mixed
+        outputs: key to store the cut-mixed input
+        mode: what mode to execute in. Probably 'train'
+        alpha: the alpha value defining the beta distribution to be drawn from during training
+        framework: which framework the current Op will be executing in. Either 'tf' or 'torch'
+    """
+    def __init__(self,
+                 inputs: Union[str, List[str]],
+                 outputs: Union[str, List[str]],
+                 mode: Union[None, str, Iterable[str]] = None,
```
TortoiseHam, line 31: can set default mode value to 'train' rather than just saying "probably train" in doc string

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,77 @@
...
+class CutMixBatch(TensorOp):
+    """This class perform cutmix augmentation on batch of tensors.
+
+    In this augmentation technique patches are cut and pasted among traning images where the ground truth labels are
+    also mixed proportionally to the area of the patches. This class should be used in conjunction with MixUpLoss to
+    perform CutMix training, which helps to reduce over-fitting, perform object detection, and against adversarial
+    attacks (https://arxiv.org/pdf/1905.04899.pdf)
+
+    Args:
+        inputs: key of the input to be cut-mixed
```
TortoiseHam, line 22: Doc string arg descriptions should be sentences (capitalized, end in '.')

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,77 @@
...
+        outputs: key to store the cut-mixed input
+        mode: what mode to execute in. Probably 'train'
+        alpha: the alpha value defining the beta distribution to be drawn from during training
+        framework: which framework the current Op will be executing in. Either 'tf' or 'torch'
+    """
+    def __init__(self,
+                 inputs: Union[str, List[str]],
+                 outputs: Union[str, List[str]],
```
TortoiseHam, line 30: outputs has to be both lambda key and img_out key, so typing should only be List[str]. this should also be clarified in doc string and an assert in the __init

TortoiseHam: or better yet, make it 'output' and 'mix_key' as separate args here, and default the 'mix_key' to 'lam' or something

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,77 @@
...
+                 mode: Union[None, str, Iterable[str]] = None,
+                 alpha: Union[float, Tensor] = 1.0,
+                 framework: str = 'tf') -> None:
```
TortoiseHam, line 33: should not take framework arg

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,77 @@
...
+            self.uniform = tfp.distributions.Uniform()
+        elif framework == 'torch':
+            self.alpha = torch.tensor(self.alpha)
+            self.beta = torch.distributions.beta.Beta(self.alpha, self.alpha)
+            self.uniform = torch.distributions.uniform.Uniform(low=0, high=1)
+        else:
+            raise ValueError("unrecognized framework: {}".format(framework))
+
+    def forward(self, data: Tensor, state: Dict[str, Any]) -> Tuple[Tensor, Tensor]:
+        """ Forward method to perform cutmix batch augmentation
```
TortoiseHam, line 53: I don't think we typically have doc strings on the forward methods unless they're doing something weird. people can read the doc string in the parent class

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,77 @@
...
+        if tf.is_tensor(data):
+            patches = roll(
+                data, shift=1,
+                axis=0)[:, bbox_y1:bbox_y2, bbox_x1:bbox_x2, :] - data[:, bbox_y1:bbox_y2, bbox_x1:bbox_x2, :]
+            patches = tf.pad(patches, [[0, 0], [bbox_y1, height - bbox_y2], [bbox_x1, width - bbox_x2], [0, 0]],
+                             mode="CONSTANT",
+                             constant_values=0)
+            data = data + patches
+            # adjust lambda to match pixel ratio
+            lam = tf.dtypes.cast(1.0 - (bbox_x2 - bbox_x1) * (bbox_y2 - bbox_y1) / (width * height), tf.float32)
```
TortoiseHam, line 72: lambda adjust should be possible outside of the if/else using our backend cast() method

fastestimator/backend/random_mix_patch.py
```diff
@@ -0,0 +1,57 @@
...
+from fastestimator.backend.cast import cast
+from fastestimator.backend.clip_by_value import clip_by_value
+from fastestimator.backend.get_image_dims import get_image_dims
+from fastestimator.backend.tensor_round import tensor_round
+from fastestimator.backend.tensor_sqrt import tensor_sqrt
+
+Tensor = TypeVar('Tensor', tf.Tensor, torch.Tensor)
+
+
+def random_mix_patch(tensor: Tensor, x: Tensor, y: Tensor, lam: Tensor) -> Tensor:
```
vbvg2008, line 29: let's move this function to where we use it.

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,71 @@
...
+    perform CutMix training, which helps to reduce over-fitting, perform object detection, and against adversarial
+    attacks (https://arxiv.org/pdf/1905.04899.pdf).
+
+    Args:
+        inputs: Key(s) of the input to be cut-mixed.
+        outputs: Key(s) under which to store the cut-mixed images and lambda value.
+        mode: What mode(s) to execute this Op in. For example, "train", "eval", "test", or "infer". To execute
+            regardless of mode, pass None. To execute in all modes except for a particular one, you can pass an argument
+            like "!infer" or "!train".
+        alpha: The alpha value defining the beta distribution to be drawn from during training.
```
vbvg2008, line 27: which controls the combination ratio between image pairs.

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    def __init__(self,
+                 inputs: Union[str, List[str]],
+                 outputs: List[str],
+                 mode: Union[None, str, Iterable[str]] = 'train',
+                 alpha: Union[float, Tensor] = 1.0) -> None:
+        assert alpha > 0, "Alpha value must be greater than zero"
+        assert len(outputs) >= 2, "Outputs should have at least two string keys"
```
vbvg2008, line 35: elaborate:  data and lambda

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -0,0 +1,71 @@
...
+        super().__init__(inputs=inputs, outputs=outputs, mode=mode)
+        self.alpha = alpha
+        self.beta = None
+        self.uniform = None
+
+    def build(self, framework: str) -> None:
+        if framework == 'tf':
+            self.alpha = tf.constant(self.alpha)
```
vbvg2008, line 43: no need to convert self.alpha to tensor

PR #822, opened 2020-08-03
fix the issue that one-layer torch model cannot be trained

geez0219:
1. fix the issue that one-layer torch model cannot be trained
2. add the corresponding testing code

### DIFF
--- a/fastestimator/backend/get_gradient.py
+++ b/fastestimator/backend/get_gradient.py
@@ -78,16 +78,16 @@ def get_gradient(target: Tensor,
                                         create_graph=higher_order,
                                         only_inputs=True)
 
-        if isinstance(gradients, Tuple) and len(gradients) == 1 and isinstance(sources, torch.Tensor):
+        if isinstance(sources, torch.Tensor):
             #  The behavior table of tf and torch backend
             #  ---------------------------------------------------------------
             #        | case 1                     | case 2                    |
             #  ---------------------------------------------------------------|
-            #  tf    | taget: tf.Tensor           | taget: tf.Tensor          |
+            #  tf    | target: tf.Tensor          | target: tf.Tensor         |
             #        | sources: tf.Tensor         | sources: [tf.Tensor]      |
             #        | gradients: tf.Tensor       | gradients: [tf.Tensor]    |
             # ----------------------------------------------------------------|
-            #  torch | taget: torch.Tensor        | taget: tf.Tensor          |
+            #  torch | target: torch.Tensor       | target: tf.Tensor         |
             #        | sources: torch.Tensor      | sources: [tf.Tensor]      |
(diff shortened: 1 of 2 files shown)


fastestimator/backend/get_gradient.py
```diff
@@ -77,8 +77,23 @@ def get_gradient(target: Tensor,
...
                                         only_inputs=True)
-        if isinstance(gradients, Tuple) and len(gradients) == 1:
-            # PyTorch seems to unnecessarily wrap results into a tuple when a non-scalar target is provided
+
+        if isinstance(gradients, Tuple) and len(gradients) == 1 and isinstance(sources, torch.Tensor):
+            #  The behavior table of tf and torch backend
+            #  ---------------------------------------------------------------
+            #        | case 1                     | case 2                    |
+            #  ---------------------------------------------------------------|
+            #  tf    | taget: tf.Tensor           | taget: tf.Tensor          |
```
TortoiseHam, line 86: 'target' rather than 'taget'

fastestimator/backend/get_gradient.py
```diff
@@ -77,8 +77,23 @@ def get_gradient(target: Tensor,
                                         retain_graph=retain_graph,
                                         create_graph=higher_order,
                                         only_inputs=True)
-        if isinstance(gradients, Tuple) and len(gradients) == 1:
-            # PyTorch seems to unnecessarily wrap results into a tuple when a non-scalar target is provided
+
+        if isinstance(gradients, Tuple) and len(gradients) == 1 and isinstance(sources, torch.Tensor):
```
TortoiseHam, line 81: do we still need to check that len(gradients) == 1 if we're checking that sources is a torch.Tensor?

fastestimator/backend/get_gradient.py
geez0219: Apparently i have a spelling issue.

fastestimator/backend/get_gradient.py
geez0219: I was considering you may see some cases that I didn't, so I just added an additional condition.
Since you also think that is unnecessary then I will simplify it.

geez0219: In that sense, I might also discard "isinstance(gradients, Tuple)" condition

TortoiseHam: so long as your code comments about possible combinations is exhaustive. have you run pytorch saliency with these changes?

geez0219: No, but i have run nightly-test and it passed

geez0219: Just ran the xai pytorch version that was modified from t08_xai.ipynb and it works

PR #825, opened 2020-08-08
Clean up mixup and cut mix 

TortoiseHam:


TortoiseHam:
this also fixes log parsing issue where it wouldn't display keys with spaces in the name, as well as handling scientific notation

PR #830, opened 2020-08-12
added option to download weights from google drive

purujitb:


fastestimator/network.py
```diff
@@ -30,6 +30,10 @@
 from fastestimator.util.traceability_util import trace_model, traceable
 from fastestimator.util.util import NonContext, get_batch_size, to_list
 
+import gdown
```
TortoiseHam, line 33: import order not following convention

fastestimator/network.py
```diff
@@ -768,8 +772,16 @@ def _fe_compile(model: Model,
         model.current_optimizer = optimizer_fn
     model.optimizer = optimizer_fn
     model.fe_compiled = True
+
     if weight:
+        if not os.path.isfile(weight):
+            tmp_dir = tempfile.mkdtemp()
```
TortoiseHam, line 778: weren't we going to combine this with the ~/.fastestimator_data directory to cache the downloads?

PR #841, opened 2020-09-09
new test report generating trace -- ModelEval 

geez0219:
1. new test report generating trace
2. integration test of that trace


fastestimator/util/latex_util.py
```diff
@@ -36,6 +36,54 @@ def dumps(self) -> str:
         return self.dumps_content()
 
 
+class TabularCell(Container):
+    """A class to represent a tabularx cell that can directly used in Tabular.add_row(). The purpose of this class is to
```
TortoiseHam, line 40: our doc-string convention is the first line needs to be short enough to not wrap to the next line. more details are given in second paragraph

fastestimator/util/latex_util.py
```diff
@@ -36,6 +36,54 @@ def dumps(self) -> str:
...
+    fix the issue that first element of Tabularx cells is not wrappable.
```
TortoiseHam, line 41: it's not clear to me what this means

fastestimator/util/latex_util.py
```diff
@@ -36,6 +36,54 @@ def dumps(self) -> str:
...
+        """Get a string representation of this cell.
+
+        Returns:
+            A string representation of itself.
+        """
+        return dumps_list(self, token=self.token)
+
+
+class WrapText(LatexObject):
+    """A class to represent a string that will be wrappable text if it is long enough. The purpose of this class is to
```
TortoiseHam, line 61: same doc string format issue

fastestimator/util/latex_util.py
```diff
@@ -36,6 +36,54 @@ def dumps(self) -> str:
...
+    def dumps(self) -> str:
+        """Get a string representation of this cell.
+
+        Returns:
+            A string representation of itself.
+        """
+        if len(self.data) > self.seq_thld:
+            return NoEscape(r'\seqsplit{' + _latex_item_to_string(self.data) + '}')
+        else:
+            return _latex_item_to_string(self.data)
```
TortoiseHam, line 84: probably don't want to be invoking a private method to do this transform. is there are reason you can't use escape_latex() method instead?

fastestimator/util/traceability_util.py
```diff
@@ -1080,3 +1085,33 @@ def fe_state(self) -> Mapping[str, Any]:
         return cls
 
     return make_traceable
+
+
+def get_environment() -> dict:
```
TortoiseHam, line 1090: revert this since no longer being used in the new trace

fastestimator/trace/io/traceability.py
```diff
@@ -438,14 +437,23 @@ def _document_models(self) -> None:
     def _document_sys_config(self) -> None:
         """Add a system config summary to the traceability document.
         """
+        env_dict = get_environment()
```
TortoiseHam, line 440: this can all be reverted since not used in new trace

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+from pylatex import Command, Document, Figure, Hyperref, Itemize, Label, LongTable, Marker, MultiColumn, NoEscape, \
+    Package, Section, Subsection, Subsubsection, Table, Tabular, Tabularx, TextColor, escape_latex
+from pylatex.utils import bold, dumps_list
+
+import fastestimator as fe
+from fastestimator.summary.logs.log_plot import visualize_logs
+from fastestimator.trace.trace import Trace
+from fastestimator.util.data import Data
+from fastestimator.util.latex_util import PyContainer, TabularCell, WrapText
+from fastestimator.util.traceability_util import get_environment, traceable
```
TortoiseHam, line 36: unused import

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+from fastestimator.util.util import to_list, to_number
+
+
+@traceable()
+class TestCase():
+    """This class defines the test case that TestReport trace will take to perform auto-testing.
+
+    Args:
+        description: A test description.
+        criteria: Function to perform the test that return True when test passes and False when test fails. Input
```
TortoiseHam, line 46: that return**s** True when **the** test passes

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+            variable name will be used as input keys to futher derive input value.
```
TortoiseHam, line 47: 'to further derive input value' ?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+        aggregate: If True, this test will be treated as per-instance type of which test criteria will be examined at
```
TortoiseHam, line 48: Why is aggregate=True per-instance?

fastestimator/trace/io/__init__.py
```diff
@@ -20,4 +20,5 @@
 from fastestimator.trace.io.qms import QMSTest
 from fastestimator.trace.io.restore_wizard import RestoreWizard
 from fastestimator.trace.io.tensorboard import TensorBoard
+from fastestimator.trace.io.model_eval import ModelEval
```
TortoiseHam, line 23: these are supposed to be sorted alphabetically

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+            self.result = []
+            self.fail_id = []
+
+
+@traceable()
+class ModelEval(Trace):
+    """Automate testing and report generation.
+
+    Args:
+        test_cases: List of TestCase object.
```
TortoiseHam, line 73: object**s**, except it's also not always a list based on your typing below

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+        save_path: Where to save the output directory
+        test_title: Title of the test.
```
TortoiseHam, line 75: is this not going to be the experiment name?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+        data_id: Data instance ID key. If provided, then sample-case test will return failing sample ID.
```
TortoiseHam, line 76: need to decide on a consistent name for these tests. is it 'sample-case', 'per-instance', or something else?

geez0219: Oops! I miss to refactor that part.

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+            with self.doc.create(Itemize()) as itemize:
+                itemize.add_item(
+                    escape_latex("Execution time: {:.2f} seconds".format(self.json_summary['execution_time(s)'])))
+
+            with self.doc.create(Table(position='H')) as table:
+                table.append(NoEscape(r'\refstepcounter{table}'))
+                self._document_summary_table(pass_num=len(instance_pass_tests) + len(aggregate_pass_tests),
+                                                fail_num=len(instance_fail_tests) + len(aggregate_fail_tests))
+
+        if len(instance_pass_tests) + len(aggregate_pass_tests) > 0:
```
TortoiseHam, line 242: could just do 

if instance_pass_tests or aggregate_pass_tests:

no need for len()

fastestimator/trace/io/model_eval.py
geez0219: the save_path is the path to saved directory like traceability does. Report will be named by experiment name with "_test" surffix

fastestimator/trace/io/model_eval.py
geez0219: Ya it is more elegant.

fastestimator/trace/io/model_eval.py
geez0219: I forgot to change the docstring description.

fastestimator/trace/io/model_eval.py
geez0219: Ya, to further derive input value

fastestimator/trace/io/model_eval.py
TortoiseHam: why have a separate test title and experiment name?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+            result = case.criteria(*[data[var_name] for var_name in case.criteria_inputs])
+            if not isinstance(result, np.ndarray):
+                raise TypeError("Criteria return of per-instance test need to be ndarray with dtype bool")
+            elif result.dtype != np.dtype("bool"):
+                raise TypeError("Criteria return of per-instance test need to be ndarray with dtype bool")
+
+            result = result.reshape(-1)
+            case.result.append(result)
+            if self.sample_id:
+                data_id = to_number(data[self.sample_id]).reshape((-1, ))
```
TortoiseHam, line 129: why are we assuming that data_id must be a number?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+                    raise ValueError("Array size of criteria return doesn't match ID array size."
+                                     "Criteria return size should be equal to the batch_size that each entry represents"
+                                     "test result of corresponding data instance")
+                case.fail_id.append(data_id[result == False])
+
+    def on_epoch_end(self, data: Data) -> None:
+        for case in self.aggregate_cases:
+            result = case.criteria(*[data[var_name] for var_name in case.criteria_inputs])
+            if not isinstance(result, (bool, np.bool_)):
+                raise TypeError("criteria return of epoch-case test need to be bool")
```
TortoiseHam, line 140: should let the user know which tests are giving the invalid response. also 'needs' rather than 'need'

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+                        with self.doc.create(Table(position='H')) as table:
+                            table.append(NoEscape(r'\refstepcounter{table}'))
+                            self._document_aggregate_table(tests=aggregate_pass_tests)
+                if len(instance_pass_tests) > 0:
+                    with self.doc.create(Subsection("Passed Per-Instance Tests")):
+                        with self.doc.create(Table(position='H')) as table:
+                            table.append(NoEscape(r'\refstepcounter{table}'))
+                            self._document_instance_table(tests=instance_pass_tests, with_ID=self.sample_id)
+
+        if len(instance_fail_tests) + len(aggregate_fail_tests) > 0:
```
TortoiseHam, line 255: might want to report failed tests before passed tests

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+
+    def _document_summary_table(self, pass_num:int, fail_num:int) -> None:
+        """Document summary table
+
+        Args:
+            pass_num: Total number of passed tests
+            fail_num: Total number of failed tests
+        """
+        with self.doc.create(Tabularx('|Y|Y|Y|', booktabs=True)) as tabular:
+            package = Package('xcolor', options='table')
```
TortoiseHam, line 276: are you using colored text?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+                if with_ID:
+                    id_data = [WrapText(data=x, seq_thld=27) for x in test["fail_id"]]
+                    row_cells.append(TabularCell(data=id_data, token=", "))
+
+                tabular.add_row(row_cells)
+                self.test_id += 1
+
+
+    def _document_aggregate_table(self, tests: List[dict]) -> None:
+        """Document aggregate table
```
TortoiseHam, line 345: table.

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+                    self.test_id,
+                    TabularCell(data=des_data, token=" "),
+                    TabularCell(data=inp_data, token=escape_latex(", \n")),
+                ]
+
+                tabular.add_row(row_cells)
+                self.test_id += 1
+
+    def _dump_pdf(self) -> None:
+        """Dump PDF file
```
TortoiseHam, line 383: file.

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+        """
+        # Need to move the tikz dependency after the xcolor package
+        self.doc.dumps_packages()
+        packages = self.doc.packages
+        tikz = Package(name='tikz')
+        packages.discard(tikz)
```
TortoiseHam, line 389: are you using tikz?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+        else:
+            # Force a double-compile since some compilers will struggle with TOC generation
+            self.doc.generate_pdf(os.path.join(self.save_dir, self.report_name), clean_tex=False, clean=False)
+            self.doc.generate_pdf(os.path.join(self.save_dir, self.report_name), clean_tex=False)
+            suffix = '.pdf'
+        print("FastEstimator-TestReport: Report written to {}{}".format(os.path.join(self.save_dir, self.report_name),
+                                                                        suffix))
+
+    def _dump_json(self) -> None:
+        """Dump JSON file
```
TortoiseHam, line 405: file.

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,446 @@
...
+
+    @staticmethod
+    def check_pdf_dependency() -> None:
+        """Check dependency of PDF-generating packages
+
+        Raises:
+            OSError: Some required package has not been installed
+        """
+        try:
+            pydot.Dot.create(pydot.Dot())
```
TortoiseHam, line 437: are you using pydot?

fastestimator/trace/io/model_eval.py
geez0219: It doesn't assume data_id to be a number. It just converts it from tensor to Numpy array. The Numpy array can still contain strings

fastestimator/trace/io/model_eval.py
geez0219: got it.

fastestimator/trace/io/model_eval.py
geez0219: I don't understand.

fastestimator/util/latex_util.py
geez0219: fixed

fastestimator/util/latex_util.py
geez0219: fixed

fastestimator/util/latex_util.py
geez0219: fixed

fastestimator/util/latex_util.py
geez0219: fixed, now it used escape_latex() instead

fastestimator/util/traceability_util.py
geez0219: fixed, now traceability.py is reverted

fastestimator/trace/io/traceability.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/__init__.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: because i think coupling the test file title and the file name is not a good idea.

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: I indeed hadn't tested with string ID, and later did find a couple of errors in that scenario.
Now it can work with string ID.

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: No it doesn't. Fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/trace/io/model_eval.py
geez0219: No it doesn't, fixed

fastestimator/trace/io/model_eval.py
geez0219: fixed

fastestimator/util/latex_util.py
```diff
@@ -129,3 +129,54 @@ def __init__(self,
             data.append(bold(escape_latex(name)) if bold_name else escape_latex(name))
         data.append(NoEscape("}}}"))
         super().__init__(data=data)
+
+
+class TabularCell(Container):
+    """A class to convert iterable to latex representation.
+
+    The data of this class can be any type. Usually it is iterable due to its capability to setup interval string.
```
TortoiseHam, line 137: If data can be any type, then the type hint in the init method is not correct. Also, "capability to setup interval string." is not very clear. Also, if the purpose of this class is to provide a string representation of a list, then why call it a TabularCell?

fastestimator/util/latex_util.py
```diff
@@ -129,3 +129,54 @@ def __init__(self,
...
+
+class WrapText(LatexObject):
+    """A class to convert string or number to wrappable latex representation.
+
+    This class will first convert the data to string, and then to wrappable latex representation if its length is too
+    long. This is to fix the issue that first element and number of Tabularx cells is not wrappable in X column type.
+
+    Args:
+        data: String data.
+        seq_thld: When length of string is above this number, the string will be converted to wrappable text by being
```
TortoiseHam, line 164: Can we call this 'threshold' instead? Also, typically in the doc strings when we are referencing one of the other arguments we would say "When the length of `data` is greater than `threshold`, the resulting string will be made wrappable" (with the angular quotes)

fastestimator/util/latex_util.py
```diff
@@ -129,3 +129,54 @@ def __init__(self,
...
+        """Get a string representation of this cell.
+
+        Returns:
+            A string representation of itself.
+        """
+        return dumps_list(self, token=self.token)
+
+
+class WrapText(LatexObject):
+    """A class to convert string or number to wrappable latex representation.
```
TortoiseHam, line 157: strings or numbers

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+from fastestimator.util.util import to_list, to_number
+
+
+@traceable()
+class TestCase():
+    """This class defines the test case that ModelEval trace will take to perform auto-testing.
+
+    Args:
+        description: A test description.
+        criteria: Function to perform the test returns True when the test passes and False when it fails. Input
```
TortoiseHam, line 43: A function to perform the test, which returns True when

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+            variable names will be used as input keys to further derive input value.
+        aggregate: If True, this test is aggregate type and its criteria function will be examined at epoch_end. If
+            False, this test is per-instance type and its test criteria function will be examined at batch_end.
+        fail_threshold: Thershold of failure instance number to judge per-instance test as failed or passed. If failure
+            number is above this value, then the test fails; otherwise it passes. It only has effect when aggregate is
```
TortoiseHam, line 48: `aggregate` vs aggregate

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+            equal to False.
+    """
+    def __init__(self, description: str, criteria: Callable, aggregate: bool = True, fail_threshold: int = 0) -> None:
+        self.description = description
+        self.criteria = criteria
+        self.criteria_inputs = inspect.signature(criteria).parameters.keys()
+        self.aggregate = aggregate
+        if not self.aggregate:
+            self.fail_threshold = fail_threshold
```
TortoiseHam, line 57: should we raise a ValueError if the user tries to provide a fail_threshold when aggregate is True?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+
+    def clean_result(self) -> None:
+        if not self.aggregate:
+            self.result = []
```
TortoiseHam, line 61: all class member variables should appear within the init method

fastestimator/trace/io/model_eval.py
TortoiseHam: can default value be None and then have it pick up the experiment name? rather than having the default just be "Test"?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+        all_inputs = set()
+        for case in self.test_cases:
+            all_inputs.update(case.criteria_inputs)
+            case.clean_result()
+            if case.aggregate:
+                self.aggregate_cases.append(case)
+            else:
+                self.instance_cases.append(case)
+
+        if self.data_id:
```
TortoiseHam, line 99: instead you could just instantiate 'all_inputs' as all_inputs = to_set(data_id)

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+        self.test_title = test_title
+        self.test_cases = to_list(test_cases)
+        self.instance_cases = []
+        self.aggregate_cases = []
+        self.data_id = data_id
+
+        all_inputs = set()
+        for case in self.test_cases:
+            all_inputs.update(case.criteria_inputs)
+            case.clean_result()
```
TortoiseHam, line 93: if you fix their init method you won't need to call this here

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+    def on_begin(self, data: Data) -> None:
+        self._sanitize_report_name()
+        self._initialize_json_summary()
+
+    def on_batch_end(self, data: Data) -> None:
+        for case in self.instance_cases:
+            result = case.criteria(*[data[var_name] for var_name in case.criteria_inputs])
+            if not isinstance(result, np.ndarray):
+                raise TypeError(f"In test with description \'{case.description}\': "
+                                "Criteria return of per-instance test needs to be ndarray with dtype bool.")
```
TortoiseHam, line 123: this is different from the doc string for TestCase

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+        report = report.split('.')[0]
+        self.save_dir = os.path.join(root_dir, report)
+        self.src_dir = os.path.join(self.save_dir, "resources")
+        os.makedirs(self.save_dir, exist_ok=True)
+        os.makedirs(self.src_dir, exist_ok=True)
+
+        super().__init__(inputs=all_inputs, mode="test")
+
+    def on_begin(self, data: Data) -> None:
+        self._sanitize_report_name()
```
TortoiseHam, line 115: should call the clean_result() here in case the user is in interpreter / Jupiter and runs multiple .test() invocations in a row

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+                case.fail_id.append(data_id[result == False])
+
+    def on_epoch_end(self, data: Data) -> None:
+        for case in self.aggregate_cases:
+            result = case.criteria(*[data[var_name] for var_name in case.criteria_inputs])
+            if not isinstance(result, (bool, np.bool_)):
+                raise TypeError(f"In test with description \'{case.description}\': "
+                                "Criteria return of aggregate-case test needs to be a bool.")
+            case.result = case.criteria(*[data[var_name] for var_name in case.criteria_inputs])
+            case.input_val = {var_name: self._to_serializable(data[var_name]) for var_name in case.criteria_inputs}
```
TortoiseHam, line 148: shouldn't create a member variable that wasn't defined in the class init

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+
+        self.json_summary["execution_time(s)"] = time() - self.json_summary["execution_time(s)"]
+
+        self._dump_json()
+        self._init_document()
+        self._document_test_result()
+        self._dump_pdf()
+
+    def _initialize_json_summary(self) -> None:
+        """Initialize json summary
```
TortoiseHam, line 177: .

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+        """
+        self.json_summary = {
+            "title": self.test_title, "timestamp": str(datetime.now()), "execution_time(s)": time(), "tests": []
+        }
+
+    def _sanitize_report_name(self) -> None:
+        """Sanitize report name and make it class attribute
```
TortoiseHam, line 184: make it a class attribute.

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+        """
+        exp_name = self.system.summary.name
+        if not exp_name:
+            raise RuntimeError("TestReport require an experiment name to be provided in estimator.fit()")
+        # Convert the experiment name to a report name (useful for saving multiple experiments into same directory)
+        report_name = "".join('_' if c == ' ' else c for c in exp_name
+                              if c.isalnum() or c in (' ', '_')).rstrip("_").lower()
+        report_name = re.sub('_{2,}', '_', report_name) + "_ModelEval"
+        self.report_name = report_name or '_ModelEval'
```
TortoiseHam, line 193: no need for the 'or' here since you're already adding "_ModelEval" in the prior line

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+
+    def _init_document(self) -> None:
+        """Initialize latex document
```
TortoiseHam, line 196: .

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+        self.doc.preamble.append(NoEscape(r'\def\seqinsert{\-}'))
+
+        # TOC
+        self.doc.append(NoEscape(r'\tableofcontents'))
+        self.doc.append(NoEscape(r'\newpage'))
+
+    def _document_test_result(self) -> None:
+        """Document test results including test summary, passed tests, and failed tests.
+        """
+        self.test_id = 1
```
TortoiseHam, line 224: this probably shouldn't be a member variable, but if it is it should be defined in init and then just re-initialized here

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+            pass_num: Total number of passed tests.
+            fail_num: Total number of failed tests.
+        """
+        with self.doc.create(Tabularx('|Y|Y|Y|', booktabs=True)) as tabular:
+            package = Package('seqsplit')
+            if package not in tabular.packages:
+                tabular.packages.append(package)
+
+            # add table heading
+            tabular.add_row(("Total test #", "Total pass #", "Total fail #"), strict=False)
```
TortoiseHam, line 286: Total Tests / Total Passed / Total Failed

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+            tabular.add_hline()
+
+            tabular.add_row((pass_num + fail_num, pass_num, fail_num), strict=False)
```
TortoiseHam, line 289: I forget, what is strict=False for?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+
+    def _document_instance_table(self, tests: List[dict], with_ID: bool):
```
TortoiseHam, line 291: Can you put typing on the dict here? List[Dict[str, TestCase]] or w/e it is

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+            table_spec = '|c|X|c|c|'
+
+        with self.doc.create(Tabularx(table_spec, booktabs=True)) as tabular:
+            package = Package('seqsplit')
+            if package not in tabular.packages:
+                tabular.packages.append(package)
+
+            # add table heading
+            row_cells = [
+                MultiColumn(size=1, align='|c|', data="Test ID"),
```
TortoiseHam, line 310: Why are these all MultiColumn size 1? Wouldn't multicolumn only be required if at least one was supposed to actually span multiple columns?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+            ]
+
+            if with_ID:
+                row_cells.append(MultiColumn(size=1, align='c|', data="Failure data instance ID"))
+
+            tabular.add_row(row_cells)
+            tabular.add_hline()
+
+            for idx, test in enumerate(tests, 1):
+                if idx > 1:
```
TortoiseHam, line 323: you could remove this if check if you remove the add_hline on line 320 right?

TortoiseHam: and in fact, then you can remove idx from the loop entirely and drop the enumerate right?

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+                    test["fail_number"]
+                ]
+                if with_ID:
+                    id_data = [WrapText(data=x, seq_thld=27) for x in test["fail_id"]]
+                    row_cells.append(TabularCell(data=id_data, token=", "))
+
+                tabular.add_row(row_cells)
+                self.test_id += 1
+
+    def _document_aggregate_table(self, tests: List[dict]) -> None:
```
TortoiseHam, line 340: typing on the dictionary

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+        Args:
+            tests: List of corresponding test dictionary to make a table.
+        """
+        with self.doc.create(Tabularx('|c|X|X|', booktabs=True)) as tabular:
+            package = Package('seqsplit')
+            if package not in tabular.packages:
+                tabular.packages.append(package)
+
+            # add table heading
+            tabular.add_row((MultiColumn(size=1, align='|c|', data="Test ID"),
```
TortoiseHam, line 352: same question about the MultiColumn

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,451 @@
...
+                             MultiColumn(size=1, align='c|', data="Test description"),
+                             MultiColumn(size=1, align='c|', data="Input value")))
+            tabular.add_hline()
+
+            for idx, test in enumerate(tests, 1):
+                if idx > 1:
```
TortoiseHam, line 358: same idx / enumerate comment

fastestimator/trace/io/model_eval.py
TortoiseHam: doesn't seem like it was fixed?

fastestimator/trace/io/model_eval.py
geez0219: that way we cannot have default fail_threshold value (or make it default to -1?)

fastestimator/trace/io/model_eval.py
geez0219: but I am afraid people will use the same TestCase instance to run other test experiments.

fastestimator/trace/io/model_eval.py
geez0219: i am going to change the docstring to:

criteria: A function to perform the test. In aggregate tests, the function needs to return True when the test
            passes and False when it fails. In per-instance tests, the function needs to returns a ndarray of bool, where
            entries show corresponding test results. (True if the test of that data instance passes; False if the test fails).

fastestimator/trace/io/model_eval.py
geez0219: add member variable in class init

fastestimator/trace/io/model_eval.py
geez0219: The test ID will be needed in four places. I think it is okay to make it a member variable based on the fact that self.doc is also a member variable.

fastestimator/trace/io/model_eval.py
geez0219: With strict = True, add_row method will check if the number of the cells of the incoming row matches with the number of table column. Because we use redefined column type Y, the tabular cannot count the column correctly, and therefore we disable this check.

fastestimator/trace/io/model_eval.py
geez0219: The tests here is a list of dict that maps string to (string, int, bool).
The dict has some keys that match TestCase attribute but it is not TestCase.

fastestimator/trace/io/model_eval.py
geez0219: this is for changing the column type only for this row.

fastestimator/trace/io/model_eval.py
geez0219: this is for changing the column type only for this row.

fastestimator/util/latex_util.py
geez0219: changed the type hint and renamed this class as IterJoin

fastestimator/trace/io/model_eval.py
```diff
@@ -0,0 +1,464 @@
...
+
+    Args:
+        description: A test description.
+        criteria: A function to perform the test. For an aggregate test, <criteria> needs to return True when the test
+            passes and False when it fails. For a per-instance test, <criteria> needs to return an ndarray of bool,
+            where entries show corresponding test results. (True if the test of that data instance passes; False if it
+            fails).
+        aggregate: If True, this test is aggregate type and its <criteria> function will be examined at epoch_end. If
+            False, this test is per-instance type and its <criteria> function will be examined at batch_end.
+        fail_threshold: Thershold of failure instance number to judge the per-instance test as failed or passed. If
```
TortoiseHam, line 49: Spelling "Thershold"

fastestimator/trace/io/model_eval.py
TortoiseHam: It could still be zero as default, just throw if user tries to set to non zero

TortoiseHam (commented):
Some spelling stuff

fastestimator/util/latex_util.py
```diff
@@ -129,3 +129,53 @@ def __init__(self,
             data.append(bold(escape_latex(name)) if bold_name else escape_latex(name))
         data.append(NoEscape("}}}"))
         super().__init__(data=data)
+
+
+class IterJoin(Container):
+    """A class to convert an iterable to a latex representation.
+
+    The data of this class can be any type. Usually it is iterable due to its capability to setup interval string.
```
TortoiseHam, line 137: 'set up'. But also, what will happen if data isn't iterable? is there then a reason to use this class? The top line of the doc string seems to contradict this. Also, the description of 'token' seems to imply that data ought to be iterable, although it's not very clear what the 'interval' of a data entry would be.

fastestimator/util/latex_util.py
```diff
@@ -129,3 +129,53 @@ def __init__(self,
...
+            A string representation of itself.
+        """
+        return dumps_list(self, token=self.token)
+
+
+class WrapText(LatexObject):
+    """A class to convert strings or numbers to wrappable latex representation.
+
+    This class will first convert the data to string, and then to wrappable latex representation if its length is too
+    long. This is to fix the issue that first string or number is not wrappable in X column type.
```
TortoiseHam, line 160: "and then to a wrappable"

"This fixes an issue which prevents the first element placed into a latex X column from wrapping correctly."

fastestimator/util/latex_util.py
```diff
@@ -129,3 +129,53 @@ def __init__(self,
...
+
+    Args:
+        data: Input data to be converted.
+        threshold: When the length of <data> is greater than <threshold>, the resulting string will be made wrappable
```
TortoiseHam, line 164: sorry, I think GitHub may have scrambled the feedback on doctstring formatting. Data and threshold should be quoted with the ` character rather than with <>

fastestimator/trace/io/model_eval.py
TortoiseHam: to clarify, if someone is looking at only the init method, they should see all of the self.members there without going to any other functions

fastestimator/trace/io/model_eval.py
TortoiseHam: interesting. is that captured in a comment anywhere?

fastestimator/util/latex_util.py
geez0219: The reason I make it to Any at the first time is that users still can put other types to it without throwing any error, but on the other hand it violate the purpose of this class.  I will change the type hint to Iteralbe.

geez0219:
Migrated most table from tabularx to longtable.
Because longtables need table header to have add_hline, need to revert the for loop code.

fastestimator/util/latex_util.py
```diff
@@ -129,3 +129,51 @@ def __init__(self,
...
+
+    This class will first convert the data to string, and then to a wrappable latex representation if its length is too
+    long. This fixes an issue which prevents the first element placed into a latex X column from wrapping correctly.
+
+    Args:
+        data: Input data to be converted.
+        threshold: When the length of `data` is greater than `threshold`, the resulting string will be made wrappable
+    """
+    def __init__(self, data: Union[str, int, float], threshold: int):
+        assert isinstance(data, (str, int, float)), "the self.data type needs to be str, int, float"
```
TortoiseHam, line 165: sorry to keep dragging this out, but the doc string needs a Raises section for the assertion error. See fe.backend.abs for an example

vbvg2008 (approved):

PR #842, opened 2020-09-11
Add Meta TensorOps

TortoiseHam:


fastestimator/network.py
```diff
@@ -124,8 +123,7 @@ def get_loss_keys(self) -> Set[str]:
         """
         loss_keys = set()
         for op in get_current_items(self.ops):
-            if isinstance(op, UpdateOp):
-                loss_keys.update(op.inputs)
+            loss_keys |= op.get_fe_loss_keys()
```
vbvg2008, line 126: I don't think it's worth it to have `get_fe_loss_keys` as base method for every Op class.

vbvg2008 (approved):

PR #844, opened 2020-09-16
fix advanced tutorial t3 due to change in Sometimes Op

vbvg2008:


TortoiseHam (approved):
Looks good. Sorry for missing it

PR #874, opened 2020-10-01
remove hyphen in text-wrapping case and add value sanitization in test report

geez0219:


fastestimator/trace/io/test_report.py
```diff
@@ -504,3 +502,10 @@ def check_pdf_dependency() -> None:
         except ValueError:
             raise OSError("Your system locale is not configured correctly. On mac this can be resolved by adding \
                 'export LC_ALL=en_US.UTF-8' and 'export LANG=en_US.UTF-8' to your ~/.bash_profile")
+
+    @staticmethod
+    def sanitize_value(value:Union[int, float]):
```
TortoiseHam, line 507: spacing and doc string not per style guidelines

TortoiseHam:
we will also need a 1.1 pr

fastestimator/trace/io/test_report.py
TortoiseHam: also return typing arrow should be there

vbvg2008 (approved):

PR #885, opened 2020-10-07
fix docstring for website rendering purpose

geez0219:


TortoiseHam (changes requested):
would it be better to modify the website parser since in python '*var' is different from 'var'?

geez0219:
> would it be better to modify the website parser since in python '*var' is different from 'var'?

PR #889, opened 2020-10-12
RestoreWizard 2.0

TortoiseHam:
update unit tests (+2 squashed commits)
Squashed commits:
[275b0062] make RestoreWizard robust against power loss while saving
[2c66bc92] improve doc string (+3 squashed commits)
Squashed commits:
[9e5e6cb4] make is_restorable consider object size in order to prevent saving in memory datasets to disk
[227a56c3] ban deterministic training mode when using RestoreWizard
[deff637d] first pass at restore everything functionality

### DIFF
--- a/fastestimator/dataset/batch_dataset.py
+++ b/fastestimator/dataset/batch_dataset.py
@@ -166,6 +166,21 @@ class BatchDataset(FEDataset):
             results = results[0]
         return results
 
+    def __getstate__(self) -> List[Dict[Any, Any]]:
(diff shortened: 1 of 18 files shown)


TortoiseHam:
Things that still need to be tested: 

1. tf.Variables (PGGAN)
2. RestoreWizard being out of order with other traces which have on_begin methods
3. Shared torch tensors (PGGAN?)
4. Restoring state within Scheduled tops/nops/traces/datasets
5. Restoring state within MetaOps
6. Restoring state within BatchDataset datasets

vbvg2008 (approved):

PR #898, opened 2020-10-23
add test for share variable between trace and op for RestoreWizard

geez0219:


apphub/image_generation/pggan/pggan_tf.py
```diff
@@ -310,6 +311,7 @@ def forward(self, data, state):
         return tf.reduce_mean(loss)
 
 
+@traceable()  # to make this trace restorable
```
TortoiseHam, line 314: Do we want to do this in our apphub?

test/PR_test/integration_test/summary/test_system.py
```diff
@@ -278,3 +285,87 @@ def instantiate_system():
...
+        system = instantiate_system()
+        var1 = system.traces[0].var1
+        system.load_state(save_path)
+
+        with self.subTest("Check both trace and tensorop variables are reloaded"):
+            self.assertEqual(var1_new_val, system.traces[0].var1.numpy())
+            self.assertEqual(var1_new_val, system.network.ops[0].var1.numpy())
+
+        with self.subTest("Check trace and tensorop variables are still shared"):
+            var1_new_val = 3
```
TortoiseHam, line 364: what is the point of setting var1_new_val here?

apphub/image_generation/pggan/pggan_tf.py
TortoiseHam: I'm thinking probably not since the app hub doesn't use the restore wizard anyways. if we do want it then we would also need to put it in the torch file

test/PR_test/integration_test/summary/test_system.py
```diff
@@ -278,3 +285,87 @@ def instantiate_system():
...
+            ])
+            system.traces.append(TestTrace(var1=var1))
+
+            return system
+
+        system = instantiate_system()
+
+        # make some change
+        var1_new_val = 2.0
+        system.traces[0].var1[0] = 2.0
```
TortoiseHam, line 348: for consistency with the tf test, the torch tensor should use t.copy_(new_val)

apphub/image_generation/pggan/pggan_tf.py
geez0219: I thought it before but it is weird enough that pytorch script didn't throw an error as tf does if i didn't put traceable decorator there

test/PR_test/integration_test/summary/test_system.py
```diff
@@ -278,3 +285,87 @@ def instantiate_system():
...
+        var1 = system.traces[0].var1
+        system.load_state(save_path)
+
+        with self.subTest("Check both trace and tensorop variables are reloaded"):
+            self.assertEqual(var1_new_val, system.traces[0].var1.numpy())
+            self.assertEqual(var1_new_val, system.network.ops[0].var1.numpy())
+
+        with self.subTest("Check trace and tensorop variables are still shared"):
+            var1_new_val = 3
+            system.traces[0].var1.copy_(torch.tensor(var1_new_val))
```
TortoiseHam, line 365: this won't be needed if you modify lines 347 & 348 to just use the single variable

TortoiseHam (approved):
LG

PR #912, opened 2020-12-03
bbox supports float numbers

gaurav272333:
Minor change to enable float values for bounding box in our visualization util


TortoiseHam (approved):
seems fine. you've done a functional test on this right?

PR #922, opened 2020-12-18
Network post-processing

TortoiseHam:


vbvg2008:
need some tests on post processing usecases

fastestimator/network.py
```diff
@@ -656,37 +699,11 @@ def transform(self, data: Dict[str, Any], mode: str, epoch: int = 1) -> Dict[str
...
-        elif isinstance(data, set):
-            return set([self._subsample_data(val, n) for val in data])
-        elif hasattr(data, "shape") and list(data.shape) and data.shape[0] > n:
-            return data[0:n]
-        else:
-            return data
+        # Distribute multi-gpu data for processing
+        strategy = tf.distribute.get_strategy()
+        if isinstance(strategy, tf.distribute.MirroredStrategy):
+            data = next(iter(strategy.experimental_distribute_dataset(tf.data.Dataset.from_tensors(data))))
```
vbvg2008, line 705: this is creating a new instance of tf.Dataset every step.    it may impact our training speed.

fastestimator/network.py
```diff
@@ -454,36 +513,20 @@ def _detach_tensor(self, data: T) -> T:
         elif isinstance(data, torch.Tensor):
             return data.detach()
 
-    def transform(self, data: Dict[str, Any], mode: str, epoch: int = 1) -> Dict[str, Any]:
```
vbvg2008, line 457: we need to see the training speed before removing the `transform`

fastestimator/network.py
TortoiseHam: no, training uses run_step, not transform()

fastestimator/network.py
TortoiseHam: transform is still there, and has nothing to do with training speed

vbvg2008 (approved):

PR #924, opened 2020-12-23
allowing use of dict as batchsize, fix potential benchmark bug

vbvg2008:
the original benchmark function was not considering use case of having scheduler as batch_size, this PR can fix it too.

### DIFF
--- a/fastestimator/pipeline.py
+++ b/fastestimator/pipeline.py
@@ -63,7 +63,7 @@ class Pipeline:
                  train_data: Union[None, DataSource, Scheduler[DataSource]] = None,
                  eval_data: Union[None, DataSource, Scheduler[DataSource]] = None,
                  test_data: Union[None, DataSource, Scheduler[DataSource]] = None,
-                 batch_size: Union[None, int, Scheduler[int]] = None,
+                 batch_size: Union[None, int, Scheduler[Union[int, Dict[str, int]]], Dict[str, int]] = None,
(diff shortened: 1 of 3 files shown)


fastestimator/pipeline.py
```diff
@@ -184,7 +184,10 @@ def benchmark(self, mode: str = "train", epoch: int = 1, num_steps: int = 1000,
 
             data_len = len(loader.dataset.dataset)
             if self.batch_size:
-                log_interval = log_interval * self.batch_size
+                batch_size = self.batch_size.get_current_value(epoch) if isinstance(self.batch_size,
```
TortoiseHam, line 187: need to separate the dictionary unpacking here so that a scheduled dictionary still works

fastestimator/pipeline.py
```diff
@@ -118,7 +118,7 @@ def _verify_dataset(self, dataset: DataSource, **kwargs) -> bool:
         if isinstance(dataset, Dataset):
             # batch_size check
             for batch_size in get_current_items(to_list(self.batch_size)):
-                assert isinstance(batch_size, int), "unsupported batch_size format: {}".format(type(batch_size))
+                assert isinstance(batch_size, (int, dict)), "unsupported batch_size format: {}".format(type(batch_size))
```
TortoiseHam, line 121: should probably also validate the dictionary contents here

TortoiseHam (approved):

PR #948, opened 2021-02-08
SuperLoss Apphub, Fancy Logging, and LabelTracker Trace

TortoiseHam:
Also fixes a restoreWizard bug where non-traceable datasets would result in all datasets getting removed from a pipeline


vbvg2008 (approved):
LG

PR #972, opened 2021-03-17
another run-of-the-mill lr scheduler

vbvg2008:
`trace=LRScheduler(model=model, lr_fn = "arc")`

`trace=LRScheduler(model=model, lr_fn=ARC(frequency=5))`

### DIFF
--- a/fastestimator/schedule/lr_shedule.py
+++ b/fastestimator/schedule/lr_shedule.py
@@ -13,6 +13,18 @@
 # limitations under the License.
 # ==============================================================================
 import math
+import os
+from collections import deque
+from pathlib import Path
+from typing import List, Tuple, Union
+
+import cv2
+import numpy as np
+import tensorflow as tf
+import wget
+
+from fastestimator.backend.zscore import zscore
+from fastestimator.util.traceability_util import traceable
 
 
 def cosine_decay(time: int,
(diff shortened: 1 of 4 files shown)


fastestimator/schedule/lr_shedule.py
```diff
@@ -59,3 +71,125 @@ def cosine_decay(time: int,
...
+        self.frequency = frequency
+        assert isinstance(self.frequency, int) and self.frequency > 0
+        self.model = self._load_arc_model()
+        self.lr_multiplier = {0: 1.618, 1: 1.0, 2: 0.618}
+        self.train_loss_one_cycle = []
+        self.eval_loss_one_cycle = []
+        self.all_train_loss = deque([None] * 3, maxlen=3)
+        self.all_eval_loss = deque([None] * 3, maxlen=3)
+        self.all_train_lr = deque([None] * 3, maxlen=3)
+        self.use_eval_loss = True
```
TortoiseHam, line 93: the model doesn't support not having eval loss right? why this boolean?

vbvg2008: it does.    It's just not tested in experiments, but it can work.  Also, in run-of-the-mill training, we wipe away val loss 50% of the time.

fastestimator/schedule/lr_shedule.py
```diff
@@ -59,3 +71,125 @@ def cosine_decay(time: int,
...
+
+    def _preprocess_train_lr(self, train_lr: List[float], missing: int) -> np.ndarray:
+        target_size = (3 - missing) * 100
+        train_lr = np.array(train_lr) / train_lr[0]
+        train_lr = cv2.resize(train_lr, (1, target_size), interpolation=cv2.INTER_NEAREST)
+        if train_lr.size < 300:
+            train_lr = np.pad(train_lr, ((300 - train_lr.size, 0), (0, 0)), mode='constant', constant_values=1.0)
+        return train_lr
+
+    def _preprocess_train_loss(self, train_loss: List[float], missing: int) -> np.ndarray:
```
TortoiseHam, line 176: can't we combine preprocess_train_loss and preprocess_val_loss?

fastestimator/trace/adapt/lr_scheduler.py
```diff
@@ -50,25 +51,53 @@ class LRScheduler(Trace):
 
     def __init__(self, model: Union[tf.keras.Model, torch.nn.Module], lr_fn: Callable[[int], float]) -> None:
         self.model = model
-        self.lr_fn = lr_fn
-        assert hasattr(lr_fn, "__call__"), "lr_fn must be a function"
-        arg = list(inspect.signature(lr_fn).parameters.keys())
-        assert len(arg) == 1 and arg[0] in {"step", "epoch"}, "the lr_fn input arg must be either 'step' or 'epoch'"
-        self.schedule_mode = arg[0]
-        super().__init__(mode="train", outputs=self.model.model_name + "_lr")
+        self.lr_fn = ARC() if lr_fn == "arc" else lr_fn
```
TortoiseHam, line 54: given that arc behaves completely differently than a regular lr_fn, we may be better off putting all of it into a single new trace rather than trying to make a hybrid trace/scheduler approach. especially since the arc api has to be invoked in such a specific order. If we definitely want it this way maybe we can come up with a more generalizable api to interact with complex schedulers?

vbvg2008: ARC is a LR schedule, so it should be used in LR scheduler trace. Otherwise we will need to worry about what happens when user use both ARC and LR scheduler trace. 

As for the generalizable API, we will worry about it when we have other scheduler similar to ARC, then we will know the requirements of designing that API.

fastestimator/schedule/lr_shedule.py
vbvg2008: preprocessing val loss is still different from train. for example, prepcoessing val loss has an extra step of filling in zero data. 

Also, keeping them separate helps with experimenting different train/val loss processing in the future.

TortoiseHam (approved):
acceptable for now

PR #973, opened 2021-03-18
add super-convergence to apphub

geez0219:


vbvg2008:
need the real acc vs step graph.

apphub/lr_controller/super_convergence/super_convergence_tf.py
```diff
@@ -0,0 +1,197 @@
+# Copyright 2019 The FastEstimator Authors. All Rights Reserved.
```
TortoiseHam, line 1: 2021

apphub/lr_controller/super_convergence/super_convergence_tf.py
```diff
@@ -0,0 +1,197 @@
...
+    x = layers.BatchNormalization(momentum=0.8)(x)
+    x = layers.LeakyReLU(alpha=0.1)(x)
+    x = layers.Conv2D(num_channel, 3, padding='same')(x)
+    x = layers.BatchNormalization(momentum=0.8)(x)
+    x = layers.LeakyReLU(alpha=0.1)(x)
+    return x
+
+
+def my_model():
+    # prep layers
```
TortoiseHam, line 64: if we're going to keep re-using this model, should we add it to the FE architecture directory?

apphub/lr_controller/super_convergence/super_convergence_tf.py
vbvg2008: nice catch!

apphub/lr_controller/super_convergence/super_convergence_tf.py
geez0219: Maybe? Is this a well-known model?

vbvg2008: I created a task in project board.

vbvg2008 (approved):

PR #974, opened 2021-03-18
added pipeline output filter

vbvg2008:


fastestimator/dataset/op_dataset.py
```diff
@@ -65,6 +67,9 @@ def __getitem__(self, index: int) -> Mapping[str, Any]:
             items = {key: np.array([item[key] for item in items]) for key in items[0]}
         else:
             forward_numpyop(self.ops, items, {'mode': self.mode})
+        if self.output_keys:
+            for key in set(items.keys()) - self.output_keys:
```
TortoiseHam, line 71: don't need to cast items.keys() to a set for this

fastestimator/dataset/op_dataset.py
```diff
@@ -35,13 +35,15 @@ class OpDataset(Dataset):
         dataset: The base dataset to wrap.
         ops: A list of ops to be applied after the base `dataset` `__getitem__` is invoked.
         mode: What mode the system is currently running in ('train', 'eval', 'test', or 'infer').
+        output_keys: What keys can be produced from pipeline. If None, all keys will be considered.
```
TortoiseHam, line 38: perhaps 'keys should be' rather than 'keys can be'

fastestimator/dataset/op_dataset.py
vbvg2008: nice, fixed

fastestimator/dataset/op_dataset.py
vbvg2008: 'can be' is more accurate.    because it contains keys that pipeline might not produce.

fastestimator/pipeline.py
```diff
@@ -321,15 +321,19 @@ def get_results(self, mode: str = "train", epoch: int = 1, num_steps: int = 1,
...
+                   shuffle: Optional[bool] = None,
+                   output_keys: Optional[Set[str]] = None) -> Union[DataLoader, tf.data.Dataset]:
         """Get a data loader from the Pipeline for a given `mode` and `epoch`.
 
         Args:
             mode: The execution mode for the loader. This can be 'train', 'eval' or 'test'.
             epoch: The epoch index for the loader. Note that epoch indices are 1-indexed.
             shuffle: Whether to shuffle the data. If None, the value for shuffle is based on mode. NOTE: This argument
                 is only used with FastEstimator Datasets.
+            output_keys: What keys can be produced from pipeline. If None, all keys will be considered.
```
TortoiseHam, line 336: do we need to specify that this only works with our op dataset an not with tf data loader?

vbvg2008: I don't think it's needed, because the "key" concept only matters when using Ops.

TortoiseHam (approved):
LG

PR #979, opened 2021-03-25
Added architecture of Wide Residual Networks

gaurav272333:


fastestimator/architecture/pytorch/wideresnet.py
```diff
@@ -0,0 +1,149 @@
+# Copyright 2019 The FastEstimator Authors. All Rights Reserved.
```
vbvg2008, line 1: 2021, same applies to other notice

vbvg2008:
need to change the pooling layer

vbvg2008 (approved):

PR #980, opened 2021-03-25
make ResNet9 an FE architecture and import it

geez0219:


vbvg2008:
Need to make ResNet9 be ale to take all input shapes

vbvg2008 (approved):

PR #991, opened 2021-04-16
make kwargs pass to sklearn function and update trace testing

geez0219:


fastestimator/trace/metric/precision.py
```diff
@@ -72,7 +79,23 @@ def on_batch_end(self, data: Data) -> None:
 
     def on_epoch_end(self, data: Data) -> None:
         if self.binary_classification:
-            score = precision_score(self.y_true, self.y_pred, average='binary')
+            score = precision_score(self.y_true, self.y_pred, average='binary', **self.kwargs)
         else:
-            score = precision_score(self.y_true, self.y_pred, average=None)
+            score = precision_score(self.y_true, self.y_pred, average=None, **self.kwargs)
```
TortoiseHam, line 84: what does it do if we just pass conflicting kwargs without sanitizing first?

fastestimator/trace/metric/recall.py
```diff
@@ -72,7 +79,23 @@ def on_batch_end(self, data: Data) -> None:
...
         if self.binary_classification:
-            score = recall_score(self.y_true, self.y_pred, average='binary')
+            score = recall_score(self.y_true, self.y_pred, average='binary', **self.kwargs)
         else:
-            score = recall_score(self.y_true, self.y_pred, average=None)
+            score = recall_score(self.y_true, self.y_pred, average=None, **self.kwargs)
         data.write_with_log(self.outputs[0], score)
+
+    @staticmethod
+    def check_kwarg(kwargs: Dict[str, Any]) -> None:
```
TortoiseHam, line 88: should these all be named 'check_kwargs' instead of kwarg?

fastestimator/trace/metric/precision.py
geez0219: I think sanitizing first is the best practice.

fastestimator/trace/metric/recall.py
geez0219: fixed

TortoiseHam (approved):
LG

PR #992, opened 2021-04-19
Added RUA to NumpyOp

gaurav272333:


fastestimator/op/numpyop/univariate/autocontrast.py
```diff
@@ -0,0 +1,51 @@
...
+        super().__init__(inputs=inputs, outputs=outputs, mode=mode)
+        self.in_list, self.out_list = True, True
+
+    def forward(self, data: List[np.ndarray], state: Dict[str, Any]) -> List[np.ndarray]:
+        return [self._apply_autocontrast(elem) for elem in data]
+
+    def _apply_autocontrast(self, data: np.ndarray) -> np.ndarray:
+        im = Image.fromarray(data)
+        im = ImageOps.autocontrast(im)
+        return np.copy(np.asarray(im))
```
TortoiseHam, line 51: why do we need the np.copy() here?

fastestimator/op/numpyop/univariate/brightness.py
```diff
@@ -0,0 +1,58 @@
...
+class Brightness(NumpyOp):
+    """Randomly change the brightness of an image.
+
+    Args:
+        inputs: Key(s) of images to be modified.
+        outputs: Key(s) into which to write the modified images.
+        mode: What mode(s) to execute this Op in. For example, "train", "eval", "test", or "infer". To execute
+            regardless of mode, pass None. To execute in all modes except for a particular one, you can pass an argument
+            like "!infer" or "!train".
+        limit: Factor range for changing brightness.
```
TortoiseHam, line 36: Can we state the min/max bounds for the arg here? is it -1 to 1?

test/PR_test/unit_test/op/numpyop/univariate/test_color.py
```diff
@@ -0,0 +1,49 @@
...
+        ]
+        cls.multi_output_shape = (28, 28, 3)
+
+    def test_single_input(self):
+        color = Color(inputs='x', outputs='x')
+        output = color.forward(data=self.single_input, state={})
+        with self.subTest('Check output type'):
+            self.assertEqual(type(output), list)
+        with self.subTest('Check output image shape'):
+            self.assertEqual(output[0].shape, self.single_output_shape)
```
TortoiseHam, line 38: these tests aren't really validating much of anything...

fastestimator/util/util.py
```diff
@@ -104,16 +104,13 @@
 
 def parse_string_to_python(val: str) -> Any:
     """Convert a string into a python object.
-
```
TortoiseHam, line 107: proposed formatting change doesn't match our other docs

fastestimator/util/util.py
```diff
@@ -185,11 +176,34 @@ def to_set(data: Any) -> Set[Any]:
     return data
 
 
+def param_to_range(
+        data: Union[int, float, Tuple[int, int], Tuple[float, float]]) -> Union[Tuple[int, int], Tuple[float, float]]:
+    """Convert a single int or float value to a tuple signifying a range.
+    ```python
+    x = fe.util.param_to_tuple(7)  # (-7, 7)
+    x = fe.util.param_to_tuple([7, 8])  # (7,8})
```
TortoiseHam, line 184: # (7,8)

fastestimator/util/util.py
```diff
@@ -185,11 +176,34 @@ def to_set(data: Any) -> Set[Any]:
...
+    x = fe.util.param_to_tuple((3.1, 4.3))  # (3.1, 4.3)
+    x = fe.util.to_set((-3.2))  # (-3.2, 3.2)
+    ```
+    Args:
```
TortoiseHam, line 188: doc-string formatting

fastestimator/op/numpyop/univariate/autocontrast.py
gaurav272333: Pytorch throws an error otherwise: The given NumPy array is not writeable, and PyTorch does not support non-writeable tensors. This means you can write to the underlying (supposedly non-writeable) NumPy array using the tensor. You may want to copy the array to protect its data or make it writeable before converting it to a tensor. 

I can use np.array(im) instead of np.copy(np.array(im)) to make it concise and have the same effect.

fastestimator/op/numpyop/univariate/brightness.py
gaurav272333: There isn't any bound like that. I will add guidance for the factor like: the factor of 0.0 gives a black image. A factor of 1.0 gives the original image.

test/PR_test/unit_test/op/numpyop/univariate/test_color.py
gaurav272333: True. I'll write some tests specific for these ops. 
However on another note, this is the trend for almost all the NumpyOps as well.

fastestimator/util/util.py
gaurav272333: I'll rectify this.

fastestimator/util/util.py
gaurav272333: Good catch. I'll correct that.

fastestimator/util/util.py
gaurav272333: I'll rectify this.

TortoiseHam (approved):
LG

PR #996, opened 2021-04-23
fix testing and debug UpdateOp for multi-gpu

geez0219:


test/PR_test/integration_test/backend/test_update_model.py
```diff
@@ -25,11 +25,15 @@ def update(x, model):
         strategy = tf.distribute.get_strategy()
         if isinstance(strategy, tf.distribute.MirroredStrategy):
             gradients = strategy.run(update, args=(x, model))
+            gradients = [x.values[0].numpy() for x in gradients]  # from PerReplica to numpy
+            # only take the first because other are the same
```
TortoiseHam, line 29: is that true before invoking a tf merge_call()?

geez0219: Ya it is true

vbvg2008 (approved):

PR #1008, opened 2021-05-18
upgrading backend, updating readme for fe 1.2.0

vbvg2008:


TortoiseHam (approved):
LG. we'll see what happens in the nightly test

PR #1009, opened 2021-05-18
fix tf api change

vbvg2008:
the `keras_model` is moved to callbacks, and the `graph` no longer accepts step as argument.

more detail in https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/summary_ops_v2.py


TortoiseHam (approved):
LG

PR #1038, opened 2021-06-08
added warning message when deleting data in pipeline dataset.  

vbvg2008:
Also unifying all warning message formats in FE


fastestimator/dataset/op_dataset.py
```diff
@@ -75,6 +77,11 @@ def finalize(self, retain: Optional[Set[str]] = None, deep_remainder: bool = Tru
                     self[key] = self.base[key]
         if retain:
             for key in self.keys() - retain:
+                if not key in self.warned:
+                    self.warned.add(key)
+                    print("FastEstimator-Warn: the key '{}' in dataset is deleted due to not being used outside of "
+                          "Pipeline, to prevent this, you can declare the key as inputs of Trace or TensorOp.".format(
+                              key))
```
TortoiseHam, line 84:
```suggestion
                    print("FastEstimator-Warn: the key '{}' is being removed since it is unused outside of the Pipeline."
                          " To prevent this, you can declare the key as an input of a Trace or TensorOp.".format(
                              key))
```

PR #1039, opened 2021-06-09
adds warning message when pruning key in dataset

vbvg2008:
The previous PR only print warning when there's intermediate variable pruned produced by Op, this PR will make sure unused variable in dataset also gets the warning.


TortoiseHam (approved):
LG

PR #1042, opened 2021-06-10
fix Duplicate Signature while running nightly-test

geez0219:


docker/stable/Dockerfile.cpu
```diff
@@ -21,8 +21,10 @@ RUN apt-get install -y \
     texlive-latex-base \
     texlive-latex-extra
 
-# upgrade essential packages
-RUN pip install --upgrade pip setuptools==51.3.3
```
vbvg2008, line 25: other 3 files are replacing notebook install, this is replacing setuptools. is it intended?

geez0219: oops wrongly delete this line. nice catch

PR #1044, opened 2021-06-15
enable eager option in estimator.fit and test

vbvg2008:
Also make the unit test print warning instead of failing if it's due to data url issue.

### DIFF
--- a/fastestimator/cli/train.py
+++ b/fastestimator/cli/train.py
@@ -16,6 +16,7 @@ import argparse
 import json
 import os
 import sys
+from ast import literal_eval
 from typing import Any, Dict, List, Optional
 
 from fastestimator import Estimator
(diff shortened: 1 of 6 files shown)


fastestimator/cli/train.py
```diff
@@ -88,12 +91,13 @@ def configure_train_parser(subparsers: argparse._SubParsersAction) -> None:
                         dest='hyperparameters_json',
                         type=str,
                         help="The path to the hyperparameters JSON file")
-    parser.add_argument('--warmup', type=str, help="Warmup setting, can be true, false or debug", default="true")
+    parser.add_argument('--warmup', type=str, help="Warmup setting, can be true or false", default="true")
```
TortoiseHam, line 94: we could switch this to boolean type rather than going thru string parsing

vbvg2008: smart

test/PR_test/unit_test/dataset/test_data.py
```diff
@@ -41,7 +41,9 @@ def test_dataset_urls(self):
             if isinstance(value, list):
                 for url in value:
                     with self.subTest('{}{} url'.format(key, url)):
-                        self.assertTrue(ping(url))
+                        if not ping(url):
+                            print("Warning!!: {} is not reachable at {}".format(key, url))
```
TortoiseHam, line 45: Is there something with more oomf than a print message that we can do here?

vbvg2008: like print("WARNING!!!!!!!!!!!!!!!") ?

TortoiseHam: lol. like something that would show up in the test summary as neither passed nor failed.

vbvg2008: The warning is in colored font now, pretty noticeable.  Didn't find anything that can show up in the end tho.

TortoiseHam (approved):
LG

PR #1046, opened 2021-06-15
fix nightly test of key error

geez0219:
It seems that even in multi-gpu environment, it sometimes raises KeyError instead of StagingError. 
The new code will pass if either type of the errors raises


TortoiseHam (approved):
LG

PR #1051, opened 2021-06-19
Add YOLOv5 Apphub

vbvg2008:


apphub/instance_detection/yolov5/yolov5_tf.py
```diff
@@ -0,0 +1,579 @@
...
+from fastestimator.op.tensorop import Average, TensorOp
+from fastestimator.op.tensorop.model import ModelOp, UpdateOp
+from fastestimator.schedule import EpochScheduler, cosine_decay
+from fastestimator.trace.adapt import LRScheduler
+from fastestimator.trace.io import BestModelSaver
+from fastestimator.trace.metric import MeanAveragePrecision
+from fastestimator.util import get_num_devices
+
+
+# This dataset selects 4 images and its bboxes
```
TortoiseHam, line 42:
```suggestion
# This dataset selects 4 images and their bboxes
```

apphub/instance_detection/yolov5/yolov5_tf.py
```diff
@@ -0,0 +1,579 @@
...
+class PreMosaicDataset(Dataset):
+    def __init__(self, mscoco_ds):
+        self.mscoco_ds = mscoco_ds
+
+    def __len__(self):
+        return len(self.mscoco_ds)
+
+    def __getitem__(self, idx):
+        indices = [idx] + [random.randint(0, len(self) - 1) for _ in range(3)]
```
TortoiseHam, line 51: does it matter that this could draw the same image multiple times? Or is probability low enough to just not bother avoiding?

apphub/instance_detection/yolov5/yolov5_tf.py
```diff
@@ -0,0 +1,579 @@
...
+                x2_new = np.clip(x1 + padw + bw, x1a, x2a)
+                y2_new = np.clip(y1 + padh + bh, y1a, y2a)
+                bw_new = x2_new - x1_new
+                bh_new = y2_new - y1_new
+                if bw_new * bh_new > 1:
+                    bboxes_new.append((x1_new, y1_new, bw_new, bh_new, label))
+        return images_new, bboxes_new
+
+
+class HSVAugment(NumpyOp):
```
TortoiseHam, line 107: Our HueSaturationValue op didn't work for this? If not, should this one replace our current offering?

apphub/instance_detection/yolov5/yolov5_tf.py
vbvg2008: the probability is 1/120k, but it's ok to draw the same image.

apphub/instance_detection/yolov5/yolov5_tf.py
```diff
@@ -0,0 +1,579 @@
...
+    out_23 = layers.Conv2D((num_classes + 5) * 3,
+                           1,
+                           bias_initializer=Constant(biases[2]),
+                           kernel_regularizer=tf.keras.regularizers.L2(0.0005),
+                           bias_regularizer=tf.keras.regularizers.L2(0.0005))(x_23)
+    out_23 = layers.Reshape((out_23.shape[1], out_23.shape[2], 3, num_classes + 5))(out_23)  # B, h/32, w/32, 3, 85
+    return tf.keras.Model(inputs=inp, outputs=[out_17, out_20, out_23])
+
+
+class ComputeLoss(TensorOp):
```
TortoiseHam, line 288: This should extend LossOp

apphub/instance_detection/yolov5/yolov5_tf.py
```diff
@@ -0,0 +1,579 @@
...
+                if diou:
+                    return iou - rho2 / c2
+                elif ciou:
+                    v = (4 / math.pi**2) * tf.pow(tf.atan(w2 / h2) - tf.atan(w1 / h1), 2)
+                    alpha = v / (1 - iou + v)
+                    return iou - (rho2 / c2 + v * alpha)
+        return tf.clip_by_value(iou, 0, 1)
+
+
+class Rescale(TensorOp):
```
TortoiseHam, line 345: Could just use the Normalize op w/ mean=0, std=1

apphub/instance_detection/yolov5/yolov5_tf.py
vbvg2008: From what I see, it looks like this implementation determines has different shift for different pixel value, rather than using the same shift throughout the image.  I think both Ops have their pros and cons, therefore no need to have one replacing the other.    This Op is used by official code, so I stick to this one for consistency.

apphub/instance_detection/yolov5/yolov5_tf.py
vbvg2008: but it'd be less intuitive for new user trying to learn what is going on. Since they need to know Normalize divides value by 255 under the hood.

apphub/instance_detection/yolov5/yolov5_tf.py
vbvg2008: done

apphub/instance_detection/yolov5/yolov5_tf.py
vbvg2008: done

TortoiseHam (approved):
LG

PR #1054, opened 2021-06-24
enable nested list in dataset and BatchDataset

vbvg2008:


fastestimator/estimator.py
```diff
@@ -340,7 +339,7 @@ def _configure_loader(self, loader: Union[DataLoader, tf.data.Dataset]) -> Union
         new_loader = loader
         if isinstance(new_loader, DataLoader) and isinstance(self.network, TFNetwork):
             add_batch = True
-            if hasattr(loader.dataset, "dataset") and isinstance(loader.dataset.dataset, BatchDataset):
+            if hasattr(loader.dataset, "dataset") and loader.dataset.dataset.fe_is_batched:
```
TortoiseHam, line 342: the member variable should be _fe_is_batched to indicate private var

vbvg2008: doesn't make it less janky.    in model, we have model.fe_compiled

TortoiseHam: slightly less janky. we should have done _fe_compiled for the model. We use ._ everywhere else that we are inserting random stuff into objects I think

TortoiseHam:
BatchDataset should have the batch flag true by default. Our concrete fe datasets can have the flag false by default. fe/trace/io/traceability.py line 539 needs updating to look at the flag instead of batchDataset instance check. Perhaps the flag could even be an integer representing the true batch size rather than a bool?

fastestimator/dataset/batch_dataset.py
```diff
@@ -224,7 +232,11 @@ def __getitem__(self, batch_idx: int) -> List[Dict[str, Any]]:
...
             for dataset, num_sample, index_map in zip(self.datasets, num_samples, self.index_maps):
                 for idx in range(num_sample):
-                    items.append(dataset[index_map[batch_idx * num_sample + idx]])
+                    item = dataset[index_map[batch_idx * num_sample + idx]]
+                    if isinstance(item, list):
+                        items.extend(item)
+                    else:
+                        items.append(item)
         else:
             num_sample = self.num_samples[0]
```
TortoiseHam, line 241: else branch doesn't need editing? Is the idea that disjoint keys have to have same size? wouldn't special handling still be needed if, for example, each disjoint dataset returned a list of 2 things?

TortoiseHam: or actually, does the implementation of _check_input that you proposed just prevent the usage of these list outputs with disjoint ds? since there'd be more than 1 copy of a key?

fastestimator/estimator.py
TortoiseHam: also, do we need a hasattr guard here?

fastestimator/pipeline.py
```diff
@@ -363,7 +367,7 @@ def get_loader(self,
                                    output_keys,
                                    deep_remainder=False)
             # Results will be immediately converted to tensors, so don't need deep_remainder
-            batch_size = None if isinstance(data, BatchDataset) else batch_size
+            batch_size = None if data.fe_is_batched else batch_size
             data = DataLoader(op_dataset,
                               batch_size=batch_size,
                               shuffle=False if isinstance(data, BatchDataset) else shuffle,
```
TortoiseHam, line 373: need to swap this is instance check to right? or maybe not due to the reset_index_maps thing? how will other datasets that are manually batching handle shuffle?

TortoiseHam: likewise for the sampler

fastestimator/estimator.py
vbvg2008: instance check it is

fastestimator/dataset/batch_dataset.py
```diff
@@ -81,7 +81,18 @@ def _check_input(self) -> None:
         for num_sample in self.num_samples:
             assert isinstance(num_sample, int) and num_sample > 0, "only accept positive integer type as num_sample"
         # check dataset keys
-        dataset_keys = [set(dataset[0].keys()) for dataset in self.datasets]
+        dataset_keys = []
+        num_examples = self.num_samples * len(self.datasets) if len(
+            self.num_samples) == 1 else [x for x in self.num_samples]
+        for idx, dataset in enumerate(self.datasets):
+            sample_data = dataset[0]
```
TortoiseHam, line 88: we should compute the fe_batch argument here since we're already incurring the cost of calling dataset[0]

fastestimator/dataset/batch_dataset.py
```diff
@@ -215,24 +226,32 @@ def __getitem__(self, batch_idx: int) -> List[Dict[str, Any]]:
...
+            for dataset, num_sample, index_map in zip(self.datasets, self.num_samples, self.index_maps):
+                single_ds_items = []
+                for idx in range(num_sample):
+                    item = dataset[index_map[batch_idx * num_sample + idx]]
+                    if isinstance(item, list):
+                        single_ds_items.extend(item)
+                    else:
+                        single_ds_items.append(item)
+                unpaired_items.append(single_ds_items)
+            items = [{k: v for d in D for k, v in d.items()} for D in zip(*unpaired_items)]
```
TortoiseHam, line 254: I think there's a style guide complaint about using capital letter variables

fastestimator/trace/io/traceability.py
```diff
@@ -535,14 +534,17 @@ def _draw_diagram(self, mode: str, epoch: int) -> pydot.Dot:
...
-        if isinstance(ds, Dataset) and not isinstance(ds, BatchDataset):
-            batch_size = self.system.pipeline.batch_size
-            if isinstance(batch_size, Scheduler):
-                batch_size = batch_size.get_current_value(epoch)
-            if isinstance(batch_size, dict):
-                batch_size = batch_size[mode]
-            if batch_size is not None:
-                batch_size = f" (Batch Size: {batch_size})"
+        if isinstance(ds, Dataset):
+            if hasattr(ds, "fe_atch") and ds.fe_batch:
```
TortoiseHam, line 538: "fe_batch"

vbvg2008: ouch

vbvg2008: done

fastestimator/dataset/batch_dataset.py
vbvg2008: done

fastestimator/dataset/batch_dataset.py
vbvg2008: done

fastestimator/dataset/batch_dataset.py
```diff
@@ -99,8 +110,15 @@ def _check_input(self) -> None:
                 assert isinstance(p, float) and p > 0, "must provide positive float for probability distribution"
         else:
             assert len(self.datasets) == len(self.num_samples), "the number of dataset must match num_samples"
-        if not self.same_feature:
-            assert len(set(self.num_samples)) == 1, "the number of samples must be the same for disjoint features"
+        # set up batch size
+        if self.same_feature:
+            if self.probability:
+                self.fe_batch = sum([n * p for n, p in zip(num_examples, self.probability)])
```
TortoiseHam, line 116:
```suggestion
                self.fe_batch = int(sum([n * p for n, p in zip(num_examples, self.probability)]))
```

TortoiseHam (approved):
LG

PR #1057, opened 2021-07-07
Loss class weights

gaurav272333:


fastestimator/backend/binary_crossentropy.py
```diff
@@ -64,6 +78,10 @@ def binary_crossentropy(y_pred: Tensor, y_true: Tensor, from_logits: bool = Fals
         ce = tf.losses.binary_crossentropy(y_pred=y_pred,
                                            y_true=tf.reshape(y_true, y_pred.shape),
                                            from_logits=from_logits)
+        if class_weights is not None:
+            sample_weights = class_weights.lookup(tf.cast(tf.reshape(y_true, ce.shape), dtype=class_weights.key_dtype))
```
TortoiseHam, line 82: this wouldn't work If they pass a regular dictionary right?

TortoiseHam: actually nvm, it's inside the tf if check

fastestimator/backend/binary_crossentropy.py
```diff
@@ -72,6 +90,11 @@ def binary_crossentropy(y_pred: Tensor, y_true: Tensor, from_logits: bool = Fals
             ce = torch.nn.BCEWithLogitsLoss(reduction="none")(input=y_pred, target=y_true.view(y_pred.size()))
         else:
             ce = torch.nn.BCELoss(reduction="none")(input=y_pred, target=y_true.view(y_pred.size()))
+
+        if class_weights is not None:
+            sample_weights = y_true.detach().clone().cpu().apply_(lambda x: class_weights.get(int(x), 1.0))
```
TortoiseHam, line 95: why in-place apply instead of just multiplying? if you detach y_true won't this mess with gradients later for training?

fastestimator/backend/categorical_crossentropy.py
```diff
@@ -64,9 +77,17 @@ def categorical_crossentropy(y_pred: Tensor, y_true: Tensor, from_logits: bool =
...
         ce = tf.losses.categorical_crossentropy(y_pred=y_pred, y_true=y_true, from_logits=from_logits)
+        if class_weights is not None:
+            sample_weights = class_weights.lookup(tf.math.argmax(y_true, axis=-1, output_type=class_weights.key_dtype))
+            ce = ce * sample_weights
     else:
         y_true = y_true.to(torch.float)
         ce = _categorical_crossentropy_torch(y_pred=y_pred, y_true=y_true, from_logits=from_logits)
+        if class_weights is not None:
+            sample_weights = torch.argmax(y_true.detach().clone().cpu(), dim=-1).to(torch.float)
+            sample_weights.apply_(lambda x: class_weights.get(int(x), 1.0))
```
TortoiseHam, line 88: same question here about the apply

fastestimator/backend/sparse_categorical_crossentropy.py
```diff
@@ -12,20 +12,22 @@
...
 Tensor = TypeVar('Tensor', tf.Tensor, torch.Tensor)
+Weight_Dict = TypeVar('Weight_Dict', tf.lookup.StaticHashTable, Dict[int, float])
 
 
 def sparse_categorical_crossentropy(y_pred: Tensor,
                                     y_true: Tensor,
                                     from_logits: bool = False,
-                                    average_loss: bool = True) -> Tensor:
+                                    average_loss: bool = True,
+                                    class_weights: Optional[Weight_Dict] = None) -> Tensor:
```
TortoiseHam, line 30: we should have end user just pass a regular dictionary regardless. we can convert it to the staticHashTable behind the scenes. Torch might also have an equivalent that we can use to get batched lookups

fastestimator/op/tensorop/loss/cross_entropy.py
```diff
@@ -68,5 +90,11 @@ def forward(self, data: List[Tensor], state: Dict[str, Any]) -> Tensor:
                     form = "sparse"
             else:
                 form = "binary"
-        loss = self.cross_entropy_fn[form](y_pred, y_true, from_logits=self.from_logits, average_loss=self.average_loss)
+
+        class_dict = self.weight_dict["tf"] if tf.is_tensor(y_pred) else self.weight_dict["torch"]
```
TortoiseHam, line 94: this can be a single thing populated during .build() rather than making a dictionary of both

fastestimator/backend/binary_crossentropy.py
gaurav272333: y_true.detach().clone() will return a new tensor with the same data. There won't be any change in y_true. Applying in-place will operate on all elements of y_true which would allow different weights even for mask-like y_true.

fastestimator/backend/sparse_categorical_crossentropy.py
gaurav272333: As it's called in forward, won't we be creating staticHashTable every batch instead of once?

fastestimator/op/tensorop/loss/cross_entropy.py
gaurav272333: I will implement this.

TortoiseHam (approved):
LG

PR #1058, opened 2021-07-07
Fix several issues of fe.build

vbvg2008:
1. fixed the issue that caused tensorflow mixed precision model isn't working on single gpu
2. fixed the issue that caused tensorflow multi-gpu training to create model twice
3. fixed the issue that caused `fe.build` to fail when defining multiple models with single `model_fn` and providing `None` as `optimizer_fn`


TortoiseHam (approved):
LG

PR #1059, opened 2021-07-08
reduce the number of train/eval steps in nightly test down to 1

vbvg2008:
I noticed there are several apphubs that take ~1min/step to execute on CPU. This ends up taking a long time unnecessarily. Now reducing the train/eval steps per epoch to 1.

Also disabling the estimator.test in nightly test because its operation is the same as eval.

### DIFF
--- a/test/apphub_scripts/NLP/imdb/run_nb.sh
+++ b/test/apphub_scripts/NLP/imdb/run_nb.sh
@@ -18,11 +18,11 @@ set -e
 example_name="imdb"
 
 # The training arguments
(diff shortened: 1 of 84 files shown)


TortoiseHam:
2 might be better than one just in case there's a difference in the second step due to tf tracing on step 1

TortoiseHam:
Or maybe we should borrow the "one two three infinity" concept from physics. It'd still save a lot of time vs 10

vbvg2008:
I think 2 is fine. changed to 2 now

TortoiseHam (approved):
LG

PR #1060, opened 2021-07-08
Fix apphub

vbvg2008:


fastestimator/op/tensorop/model/model.py
```diff
@@ -49,6 +50,11 @@ def __init__(self,
         self.model = model
         self.trainable = trainable
         self.epoch_spec = None
+        self.multi_inputs = False
+
+    def build(self, framework: str, device: Optional[torch.device] = None) -> None:
+        if framework == "torch":
+            self.multi_inputs = len(inspect.signature(self.model.forward).parameters.keys()) > 1
```
TortoiseHam, line 57: should we also check that self has multiple inputs? it could be that the model forward has an optional parameter that the user here is choosing not to pass

vbvg2008: like this?

TortoiseHam (approved):
LG

PR #1061, opened 2021-07-09
fix apphub fails in nightly test, p2

vbvg2008:
several issues found:

1. torch multi-gpu training will create a new instance on top of existing model instance, and its forward function signature is different. Therefore, for pytorch multi-gpu model, we need to check the model.module.forward instead.

2. it seems like recent change of using *args in `feed_forward` isn't working entirely well with pytorch multi-gpu. For example, in tensorOp, "cuda:0" is now required instead of simply "cuda".   In the future, all pytorch tensorOp should use nn.Module instead, it will improve the parallel training performance.

3. found one small mistake in randaugment notebook argument, now fixed


TortoiseHam (approved):
LG. Is that really the only place we were hardcoding "cuda"?

PR #1092, opened 2021-08-17
Rua multivariate

gaurav272333:


fastestimator/op/numpyop/univariate/rua.py
```diff
@@ -180,31 +180,35 @@ def __init__(self,
...
         else:
-            self.num_bits = round(8-(magnitude_coef*self.num_bits))
+            self.num_bits = Posterize._range_tuple(num_bits=self.num_bits, magnitude_coef=magnitude_coef)
         super().__init__(inputs=self.inputs,
                          outputs=self.outputs,
                          mode=self.mode,
                          num_bits=self.num_bits)
 
+    @staticmethod
+    def _range_tuple(num_bits: Union[int, Tuple[int, int]], magnitude_coef: float) -> Tuple[int, int]:
```
TortoiseHam, line 203: we should have doc string for all non-override methods

fastestimator/op/numpyop/univariate/rua.py
```diff
@@ -255,6 +259,46 @@ def _apply_solarize(data: np.ndarray, threshold: int) -> np.ndarray:
...
+            for out in op.outputs:
+                outputs.add(out)
+
+        NumpyOp.__init__(self, inputs=inputs.union(outputs), outputs=outputs, mode=mode)
+        self.ops = numpy_ops
+
+    def forward(self, data: List[np.ndarray], state: Dict[str, Any]) -> List[np.ndarray]:
+        op_idx = random.randint(0, len(self.ops)-1)
+        data = {key: elem for key, elem in zip(self.inputs, data)}
+        forward_numpyop([self.ops[op_idx]], data, state)
```
TortoiseHam, line 298: why not just random.choice(self.ops) rather than going through the idx generation?

fastestimator/op/numpyop/univariate/rua.py
```diff
@@ -255,6 +259,46 @@ def _apply_solarize(data: np.ndarray, threshold: int) -> np.ndarray:
...
+            assert self.out_list == op.out_list, "All ops within OneOf must share the same output configuration"
+            assert mode == op.mode, "All ops within a OneOf must share the same mode"
+
+            for inp in op.inputs:
+                inputs.add(inp)
+
+            for out in op.outputs:
+                outputs.add(out)
+
+        NumpyOp.__init__(self, inputs=inputs.union(outputs), outputs=outputs, mode=mode)
```
TortoiseHam, line 292: I guess this is fine, but let's put a code comment explaining why we are doing this so we remember in future

fastestimator/op/numpyop/univariate/rua.py
gaurav272333: True. I'll change this.

fastestimator/op/numpyop/univariate/rua.py
gaurav272333: Sure. I'll add that.

fastestimator/op/numpyop/univariate/rua.py
gaurav272333: I'll add the comment. Also, if we are fine with codacy's issue, I'll update this to below code (using super instead of NumpyOp's __init__) as that seems better to me:
super(OneOf, self).__init__(inputs=inputs.union(outputs), outputs=outputs, mode=mode)

TortoiseHam: sure

TortoiseHam (approved):
LG

PR #1169, opened 2021-11-22
new data length feature introduced

DwijayDS:
Introduction of Custom sampler. One can now use "train_steps_per_epoch" or "eval_steps_per_epoch" to extend or contract the number of steps per epoch.
- Refactoring of "max_train_steps_per_epoch" and "max_eval_steps_per_epoch" to "train_steps_per_epoch" and "eval_steps_per_epoch" respectively.
- Optimization of Sampler code


TortoiseHam (commented):
The apphub Jupiter notebooks will also need to change from max_train... to train...

fastestimator/estimator.py
```diff
@@ -75,8 +75,8 @@ def __init__(self,
                  pipeline: Pipeline,
                  network: BaseNetwork,
                  epochs: int,
-                 max_train_steps_per_epoch: Optional[int] = None,
-                 max_eval_steps_per_epoch: Optional[int] = None,
+                 train_steps_per_epoch: Optional[int] = None,
```
TortoiseHam, line 78: doc string needs to be adjusted to reflect change in name/behavior

fastestimator/summary/system.py
```diff
@@ -87,8 +87,8 @@ class System:
     network: BaseNetwork
     pipeline: Pipeline
     traces: List[Union['Trace', Scheduler['Trace']]]
-    max_train_steps_per_epoch: Optional[int]
-    max_eval_steps_per_epoch: Optional[int]
+    train_steps_per_epoch: Optional[int]
```
TortoiseHam, line 90: doc string needs adjusting

fastestimator/pipeline.py
```diff
@@ -401,6 +403,8 @@ def __call__(self,
                  epoch: int = 1,
                  ds_id: Optional[str] = None,
                  shuffle: Optional[bool] = None,
+                 train_steps_per_epoch: Optional[int] = None,
```
TortoiseHam, line 406: doesn't make sense to have 'train_steps_per_epoch' and 'eval_steps_per_epoch' here since this method is specific to a given mode already. Maybe just a single arg 'steps' and have the estimator decide which argument to pass given the context

fastestimator/pipeline.py
```diff
@@ -482,6 +488,31 @@ def __enter__(self) -> Union[DataLoader, tf.data.Dataset]:
             collate_fn = self.collate_fn
             if collate_fn is None and self.pad_value is not None:
                 collate_fn = self._pad_batch_collate
+            # Default ExapandDataset Sampler
+            if self.ctx_train_steps_per_epoch is not None and self.ctx_mode == "train":
```
TortoiseHam, line 492: this if branch can be cut based on earlier recommendation

fastestimator/pipeline.py
```diff
@@ -482,6 +488,31 @@ def __enter__(self) -> Union[DataLoader, tf.data.Dataset]:
...
+
+                if self.ctx_shuffle == True:
```
TortoiseHam, line 494: instead of this if branch, just set shuffle=self.ctx_shuffle when constructing the sampler

fastestimator/pipeline.py
```diff
@@ -519,3 +551,57 @@ def _pad_batch_collate(self, batch: List[MutableMapping[str, Any]]) -> Dict[str,
         """
         pad_batch(batch, self.pad_value)
         return default_collate(batch)
+
+
+class ExtendDatasetSampler(Sampler):
+    """Samples elements randomly, while ensuring change in size of dataset.
+    Arguments:
```
TortoiseHam, line 558: doc string formatting

fastestimator/pipeline.py
```diff
@@ -519,3 +551,57 @@ def _pad_batch_collate(self, batch: List[MutableMapping[str, Any]]) -> Dict[str,
...
+        ds_len : Length of original dataset.
+        ds_expand_len : Length to which original dataset must be expanded to.
+        shuffle : Whether to use shuffling.
+    """
+    def __init__(self, ds_len, ds_expand_len, shuffle=True):
```
TortoiseHam, line 563: add typing to function arguments

fastestimator/pipeline.py
```diff
@@ -519,3 +551,57 @@ def _pad_batch_collate(self, batch: List[MutableMapping[str, Any]]) -> Dict[str,
...
+            self.indices = base[0]
+        else:
+            if self.shuffle == True:
+                random.shuffle(base[-1])
+            base[-1] = base[-1][:self.ds_expand_len % self.ds_len]
+
+            for ls in base:
+                if self.shuffle == True:
+                    random.shuffle(ls)
+                for ele in ls:
```
TortoiseHam, line 604: this can be replaced with self.indices.extend(ls)

fastestimator/pipeline.py
```diff
@@ -519,3 +551,57 @@ def _pad_batch_collate(self, batch: List[MutableMapping[str, Any]]) -> Dict[str,
...
+        """
+        assert isinstance(self.ds_expand_len, int), "Only accept positive integer type as ds_expand_len"
+        assert self.ds_expand_len > 0, "Invalid ds_expand_len. Expand Length cannot be less than or equal to 0"
+        assert isinstance(self.ds_len, int), "Only accept positive integer type as ds_len"
+        assert self.ds_len > 0, "Invalid ds_len. Dataset Length cannot be less than or equal to 0"
+
+    def __iter__(self):
+
+        self.indices = []
+        base = self.base
```
TortoiseHam, line 590: this is at best a shallow copy, so won't protect the self.base from getting shuffled

fastestimator/pipeline.py
```diff
@@ -519,3 +551,57 @@ def _pad_batch_collate(self, batch: List[MutableMapping[str, Any]]) -> Dict[str,
...
+
+        if self.ds_expand_len == self.ds_len:
```
TortoiseHam, line 592: this if condition isn't really necessary since that special case should work just fine inside the else block

fastestimator/pipeline.py
DwijayDS: The shallow copy was done to protect the variable 'self.base' from getting sliced by the line below
image

TortoiseHam: sure, but either this class is single use (in which case it doesn't need protection) or it is multi use (in which case you'd need to shield self.base from the shuffle)

TortoiseHam: ah, I guess since shuffle can't be changed after init this isn't true. this is fine as is then

fastestimator/pipeline.py
DwijayDS: In else part, this slicing line might produce an empty array if both the lengths are equal. It might make the array empty and return nothing.

image

TortoiseHam: ah, true. it may be better to use an if ds_expand_len % self.ds_len != 0 for the slicing instead though. The advantage being that less code is replicated between branches in case we add more features in future

fastestimator/estimator.py
```diff
@@ -215,10 +215,18 @@ def _warmup(self, eager: bool = True) -> None:
                         if idx > 0:  # ignore TrainEssential and EvalEssential's inputs for unmet requirement checking
                             trace_input_keys.update(trace.inputs)
                         trace_output_keys.update(trace.get_outputs(ds_ids=ds_ids))
+                    # Mode checking
+                    if mode == "train":
```
TortoiseHam, line 219: rather than duplicating this mode check code let's modify system to have:

`
@ property
def steps_per_epoch(self):
  if self.mode == 'train':
     return self.train_steps_per_epoch
  ...
`

fastestimator/estimator.py
```diff
@@ -384,10 +400,10 @@ def _configure_loader(self, loader: Union[DataLoader, tf.data.Dataset]) -> Union
             new_loader = tf.data.Dataset.from_generator(lambda: loader, data_type, output_shapes=data_shape)
             new_loader = new_loader.prefetch(1)
         if isinstance(new_loader, tf.data.Dataset):
-            if self.system.max_train_steps_per_epoch and self.system.mode == "train":
-                new_loader = new_loader.take(self.system.max_train_steps_per_epoch)
-            if self.system.max_eval_steps_per_epoch and self.system.mode == "eval":
-                new_loader = new_loader.take(self.system.max_eval_steps_per_epoch)
+            if self.system.train_steps_per_epoch and self.system.mode == "train":
+                new_loader = new_loader.take(self.system.train_steps_per_epoch)
```
TortoiseHam, line 404: what will happen here if the user tries to expand a tf.dataset that they are providing (instead of using one of our FE datasets)?

fastestimator/pipeline.py
TortoiseHam: Formatting is still off

TortoiseHam (changes requested):
The notebook and apphub integration test files will need to updated to reflect the changed argument names

fastestimator/pipeline.py
```diff
@@ -482,6 +488,17 @@ def __enter__(self) -> Union[DataLoader, tf.data.Dataset]:
             collate_fn = self.collate_fn
             if collate_fn is None and self.pad_value is not None:
                 collate_fn = self._pad_batch_collate
+            # Default ExapandDataset Sampler
+            if self.ctx_steps_per_epoch is not None and self.ctx_mode in ["train", "eval"]:
```
TortoiseHam, line 492: no need to check ctx_mode here. maybe they only want to use 3 steps of test data...

fastestimator/estimator.py
```diff
@@ -106,6 +106,16 @@ def network(self) -> BaseNetwork:
     def traces(self) -> List[Union[Trace, Scheduler[Trace]]]:
         return self.system.traces
 
+    @ property
+    def steps_per_epoch(self,mode):
```
TortoiseHam, line 110: this property can be inside the system object so that you don't have to pass the mode in order to query it

fastestimator/estimator.py
```diff
@@ -325,9 +338,13 @@ def _run_epoch(self, eager: bool) -> None:
                                     ds_id=self.system.ds_id,
                                     output_keys=trace_input_keys,
                                     eager=eager)
+            # Mode checking
+            steps = self.steps_per_epoch(mode=self.system.mode)
+
             with self.pipeline(mode=self.system.mode,
                                epoch=self.system.epoch_idx,
                                ds_id=self.system.ds_id,
+                               steps_per_epoch=steps,
```
TortoiseHam, line 347: steps_per_epoch=self.system.steps_per_epoch

fastestimator/estimator.py
DwijayDS: It might cause an issue here right? Will this property be able to work as expected here?
image

fastestimator/estimator.py
```diff
@@ -53,14 +53,19 @@ class Estimator:
     (estimator.fit) or test (estimator.test) models. It wraps `Pipeline`, `Network`, `Trace` objects together and
     defines the whole optimization process.
 
+    If the data fed into pipeline is a TensorFlow Dataset, then the parameters "train_steps_per_epoch" and
```
TortoiseHam, line 56:
```suggestion
    If the data fed into pipeline is a TensorFlow Dataset, then the parameters `train_steps_per_epoch` and
```

fastestimator/estimator.py
```diff
@@ -53,14 +53,19 @@ class Estimator:
...
+    "eval_steps_per_epoch" can only reduce the number of steps per epoch. If these parameters are higher than the
```
TortoiseHam, line 57:
```suggestion
    `eval_steps_per_epoch` can only reduce the number of steps per epoch. If these parameters are higher than the
```

fastestimator/summary/system.py
```diff
@@ -115,14 +119,23 @@ def __init__(self,
...
+        self.train_steps_per_epoch = train_steps_per_epoch
+        self.eval_steps_per_epoch = eval_steps_per_epoch
         self.stop_training = False
         self.summary = Summary(None, system_config)
         self.experiment_time = ""
         self.custom_graphs = {}
         self._initialize_state()
 
+    @property
+    def steps_per_epoch(self):
```
TortoiseHam, line 131:
```suggestion
    def steps_per_epoch(self) -> Optional[int]:
```

TortoiseHam (approved):
LG

PR #1173, opened 2021-12-01
ExtendDataset introduced

DwijayDS:


fastestimator/dataset/batch_dataset.py
```diff
@@ -287,3 +290,6 @@ def reset_index_maps(self, seed: Optional[int] = None) -> None:
                 else:
                     random.shuffle(mapping)
             self.index_maps.append([item for sublist in index_map for item in sublist])
+
+
+from fastestimator.dataset.extend_dataset import ExtendDataset
```
TortoiseHam, line 295: what is this doing down here?

fastestimator/dataset/extend_dataset.py
```diff
@@ -0,0 +1,55 @@
...
+import math
+
+from torch.utils.data import Dataset
+from fastestimator.dataset.batch_dataset import BatchDataset
+
+from fastestimator.util.traceability_util import traceable
+
+
+@traceable()
+class ExtendDataset(Dataset):
```
TortoiseHam, line 24: the dataset _init.py file should import this

fastestimator/dataset/extend_dataset.py
```diff
@@ -0,0 +1,55 @@
...
+    """ExtendDataset either extends or contracts the length of provided Dataset.
+
+    Args:
+        dataset: The Original dataset(s) which need expansion or contraction.
+        spoof_length: Length to which original dataset must be expanded or contracted to. (New desired length)
+    """
+    def __init__(self, dataset: Dataset, spoof_length: int) -> None:
+
+        self.dataset = dataset
```
TortoiseHam, line 33: let's prevent ppl from nesting extendDatasets inside one another

TortoiseHam (changes requested):
let's have a few unit tests demonstrating this works

fastestimator/dataset/batch_dataset.py
DwijayDS: It was creating an issue of circular import, which has this one solution.

fastestimator/dataset/extend_dataset.py
DwijayDS: Its prevented in "checkinfo" method

fastestimator/dataset/batch_dataset.py
DwijayDS: Solved this issue after introducing ExtendDataset in _init_.py file

test/PR_test/unit_test/dataset/test_extend_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+def inputs():
+    while True:
+        yield {'x': np.random.rand(16), 'y': np.random.randint(16)}
+
+
+class TestBatchDataset(unittest.TestCase):
+    def test_dataset_contraction(self):
+        ds1 = GeneratorDataset(generator=inputs(), samples_per_epoch=10)
+        unpaired_ds = fe.dataset.ExtendDataset(dataset=ds1, spoof_length=5)
+        pipeline = fe.Pipeline(unpaired_ds)
```
TortoiseHam, line 32: what happens when a dataset has, for example, 23 data points, batch size 25, drop_last=True and then the user extends the ds?

DwijayDS: 'y': tensor([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16, 17,18, 19, 20, 21, 22,  0,  1])

This is the result of a dataset with 23 samples and batch size of 25, with drop_last=True and spoof_length of 30

TortoiseHam (approved):
LG

PR #1178, opened 2021-12-15
Added support for multiGPU

DwijayDS:


fastestimator/backend/l2_regularization.py
```diff
@@ -38,9 +38,10 @@ def l2_regularization(model: Union[tf.keras.Model, torch.nn.Module], beta: float
         ValueError: If `model` belongs to an unacceptable framework.
     """
     if isinstance(model, torch.nn.Module):
-        l2_loss = torch.tensor(0.)
+        l2_loss = torch.tensor(0.).to(torch.device(next(model.parameters()).device))
```
TortoiseHam, line 41: to you need to wrap the device in a new torch.device()? seems like that could be skipped

fastestimator/backend/l2_regularization.py
```diff
@@ -38,9 +38,10 @@ def l2_regularization(model: Union[tf.keras.Model, torch.nn.Module], beta: float
...
         for param in model.parameters():
             if param.requires_grad:
+
```
TortoiseHam, line 44:
```suggestion
```

fastestimator/backend/l2_regularization.py
DwijayDS: I could use tensor.new(), or 
I could create tensor on the device similiar to other parameters rather than transfering them

TortoiseHam: for example, the following should work:
`a=Tensor1
 b=Tensor2
 b.to(a.device)
`
rather than:
`a=Tensor1
 b=Tensor2
 b.to(torch.device(a.device))
`

TortoiseHam (approved):
LG

PR #1185, opened 2021-12-27
improved solov2 implementation and matching official performance

vbvg2008:
* implemented the TODO item
* less regularization effects
* and train a bigger input size 
* train 36 epochs

After that, it matches with official results


TortoiseHam (approved):
LG

PR #1201, opened 2022-01-18
fix momentum value for torch apphub and architectures

vbvg2008:
the definition of BN's momentum is different between tf and torch.

if momentum is `x` in tf, then in pytorch it should be `1-x`

for more info:

https://www.tensorflow.org/api_docs/python/tf/keras/layers/BatchNormalization
https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm2d.html


TortoiseHam (approved):
watch this break everything somehow...

PR #1202, opened 2022-01-19
Updating torch model saving

aashokvardhan:


fastestimator/backend/load_model.py
```diff
@@ -65,11 +70,33 @@ def load_model(model: Union[tf.keras.Model, torch.nn.Module], weights_path: str,
...
             assert os.path.exists(optimizer_path), "cannot find optimizer path: {}".format(optimizer_path)
             model.current_optimizer.load_state_dict(torch.load(optimizer_path))
     else:
         raise ValueError("Unrecognized model instance {}".format(type(model)))
+
+
+def preprocess_torch_weights(weights_path: str) -> OrderedDict:
+    """Preprocess the torch weights dictionary.
+
+    This method is used to remove the any DataParallel artifacts in torch weigths :
```
TortoiseHam, line 89:
```suggestion
    This method is used to remove the any DataParallel artifacts in torch weigths.
```

fastestimator/backend/load_model.py
```diff
@@ -65,11 +70,33 @@ def load_model(model: Union[tf.keras.Model, torch.nn.Module], weights_path: str,
...
+
+    Args:
+        weights_path: Path to the `model` weights.
```
TortoiseHam, line 92:
```suggestion
        weights_path: Path to the model weights.
```

TortoiseHam: the back ticks are when referring to args in the method signature

fastestimator/backend/load_model.py
```diff
@@ -65,11 +70,33 @@ def load_model(model: Union[tf.keras.Model, torch.nn.Module], weights_path: str,
...
+    """
+    new_state_dict = OrderedDict()
+    for key, value in torch.load(weights_path, map_location='cpu' if torch.cuda.device_count() == 0 else None).items():
+        # remove `module.`
+        new_key = key
+        if key.startswith('module.'):
+            new_key = key[7:]
+        new_state_dict[new_key] = value
+
+    return new_state_dict
```
TortoiseHam, line 102: newline at EOF

tutorial/advanced/t15_finetuning.ipynb
```diff
@@ -663,7 +654,11 @@
     "        x = fn.relu(self.backbone_layers[2](x))\n",
     "        return x\n",
     "\n",
-    "backbone_torch = fe.build(model_fn=lambda: BackboneTorch(model_torch_pretrained), optimizer_fn=\"adam\") "
```
TortoiseHam, line 666: presumably this BackBoneTorch method should have been deleted from somewhere right?

TortoiseHam: oh, nvm

tutorial/advanced/t15_finetuning.ipynb
```diff
@@ -663,7 +654,11 @@
...
+    "\n",
+    "if isinstance(model_torch_pretrained, torch.nn.DataParallel):\n",
+    "    backbone_torch = fe.build(model_fn=lambda: BackboneTorch(model_torch_pretrained.module), optimizer_fn=\"adam\") \n",
```
TortoiseHam, line 659: is there a way we can do the backend function where the end user doesn't have to make this if-check?

aashokvardhan: We can do it if we don't use the fe.backend.load_model function in the previous step.

tutorial/advanced/t15_finetuning.ipynb
```diff
@@ -610,22 +612,11 @@
...
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
-    "model_torch_pretrained = lenet_torch(input_shape=torch_input_shape, classes=base_num_classes)\n",
+    "model_torch_pretrained = fe.build(model_fn=lambda: lenet_torch(input_shape=torch_input_shape, classes=base_num_classes), optimizer_fn=\"adam\")\n",
     "\n",
-    "model_torch_pretrained.load_state_dict(load(torch_pretrained_model_path))"
+    "fe.backend.load_model(model_torch_pretrained, weights_path=os.path.join(model_dir, 'lenet_torch.pt'))"
```
TortoiseHam: we can just put the weights path into the fe.build() call to avoid having 2 lines right?

tutorial/advanced/t15_finetuning.ipynb
TortoiseHam: why did it used to work without this consideration? or was it an existing bug?

tutorial/advanced/t15_finetuning.ipynb
aashokvardhan: Yes, we can directly put weights path into fe.build().

tutorial/advanced/t15_finetuning.ipynb
aashokvardhan: @TortoiseHam 
1) Previous implementation directly loads the weight file:
model_torch_pretrained = lenet_torch(input_shape=torch_input_shape, classes=base_num_classes)
model_torch_pretrained.load_state_dict(load(torch_pretrained_model_path))
This would fail, if the weights file previously saved from multi gpu machine directly without using fe.backend.save_model(current implementation) which would introduce "module" prefix into the keys.

2) To avoid this we are trying to use fe.backend.load_model which handles the module inconsistency, but prerequisite is it should be a fe.build.

3) But using fe.build model on multi gpu machine would generate a dataparallel module, which would create a new problem while modifying the model architecture for fine tuning, because of which we need to add if condition to handle that.

TortoiseHam: maybe this if check would be better to put inside the BackboneTorch class init since that would show the better practice for anyone who was designing based on the tutorial

tutorial/advanced/t15_finetuning.ipynb
```diff
@@ -617,15 +608,17 @@
...
      "output_type": "execute_result"
     }
    ],
    "source": [
+    "weights_path=os.path.join(model_dir, 'lenet_torch.pt')\n",
+    "\n",
     "model_torch_pretrained = lenet_torch(input_shape=torch_input_shape, classes=base_num_classes)\n",
     "\n",
-    "model_torch_pretrained.load_state_dict(load(torch_pretrained_model_path))"
+    "model_torch_pretrained.load_state_dict(load(weights_path, map_location='cpu' if cuda.device_count() == 0 else None))"
```
TortoiseHam, line 621: suppose I train on 2 gpu machine and then load weights on 1 gpu machine. would this still work? seems like this only catches the case going from multi-gpu down to cpu

tutorial/advanced/t15_finetuning.ipynb
TortoiseHam: let's show that off then

TortoiseHam (approved):
LG

PR #1204, opened 2022-01-20
 Fancier EigenCam

TortoiseHam:
improve eigencam so it can downsize activations as well as determine n_components dynamically for each sample based on captured variance


vbvg2008 (approved):
LGTM

PR #1228, opened 2022-03-01
Updating the read mat file

aashokvardhan:


fastestimator/op/numpyop/multivariate/read_mat.py
```diff
@@ -28,8 +28,9 @@ class ReadMat(NumpyOp):
     This expects every sample to have a separate .mat file.
 
     Args:
-        file: Dictionary key that contains the .mat path.
-        keys: Key(s) to read from the .mat file.
+        inputs: location of the .mat path.
+        outputs: keys to map the read data to the output.
+        keys: Keys to read from the .mat file.
```
TortoiseHam, line 33: Keys should only be required if different from outputs. also, maybe 'mat_keys' to make the distinction more obvious

apphub/multi_task_learning/uncertainty_weighted_loss/uncertainty_loss.ipynb
```diff
@@ -19,7 +19,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 10,
```
TortoiseHam, line 22: we typically restart the kernel so that the execution count doesn't display weird for people

fastestimator/op/numpyop/multivariate/read_mat.py
```diff
@@ -38,16 +39,30 @@ class ReadMat(NumpyOp):
...
         self.parent_path = parent_path
+
+        if mat_keys is None:
+            self.mat_keys = self.outputs
+        else:
+            self.mat_keys = mat_keys
+
         self.out_list = True
 
+        if isinstance(self.mat_keys, List) and isinstance(self.outputs, List):
```
TortoiseHam, line 58: this is instance is brittle is user provides keys as tuple. I'd suggest to_list on both args first

apphub/multi_task_learning/uncertainty_weighted_loss/uncertainty_loss.ipynb
TortoiseHam: sorry for the confusion, I didn't mean not to run the notebook at all. I was trying to say to restart the kernel + run the notebook such that the cells are executed in order starting from 1

fastestimator/op/numpyop/multivariate/read_mat.py
```diff
@@ -28,8 +28,9 @@ class ReadMat(NumpyOp):
     This expects every sample to have a separate .mat file.
 
     Args:
-        file: Dictionary key that contains the .mat path.
-        keys: Key(s) to read from the .mat file.
+        inputs: Dictionary key that contains the .mat path.
+        outputs: keys to map the read data to the output.
+        mat_keys: Keys to read from the .mat file. (Optional: Provide mat_keys when they are different from outputs).
```
TortoiseHam, line 33:
```suggestion
        mat_keys: (Optional) Keys to read from the .mat file. Defaults to `outputs`, but to re-name keys you can provide the original name here and the new name in `outputs`.
```

fastestimator/op/numpyop/multivariate/read_mat.py
```diff
@@ -38,16 +39,30 @@ class ReadMat(NumpyOp):
...
+            self.mat_keys = self.outputs
+        else:
+            self.mat_keys = mat_keys
+
         self.out_list = True
 
+        if isinstance(self.mat_keys, List) and isinstance(self.outputs, List):
+            assert len(self.mat_keys) == len(self.outputs), "keys and Output lengths must match"
+
+        if not isinstance(self.mat_keys, List):
```
TortoiseHam, line 61: Since doing this anyways, move it above the is instance check earlier and use to_list method instead of [] (handles None better)

TortoiseHam (approved):
LG

PR #1231, opened 2022-03-07
Update fuse

aashokvardhan:


test/PR_test/integration_test/op/numpyop/meta/test_fuse.py
```diff
@@ -41,6 +47,14 @@ def instantiate_system():
 
         self.assertEqual(loaded_var, new_var)
 
+    def test_delete_op(self):
+        ops = fe.op.numpyop.meta.Fuse(
+            ops=[TestNumpyOp(inputs='x', outputs='x', mode="train", var=1), Delete(
+                keys='x',
+                mode="train",
+            )])
+        _ = ops.forward(data=[self.tf_data], state={})
```
vbvg2008, line 56: need to assert that key `x` is indeed deleted

vbvg2008, line 56: need to assert that key `x` is indeed deleted

test/PR_test/integration_test/op/numpyop/meta/test_fuse.py
```diff
@@ -41,6 +47,14 @@ def instantiate_system():
 
         self.assertEqual(loaded_var, new_var)
 
+    def test_delete_op(self):
+        ops = fe.op.numpyop.meta.Fuse(
+            ops=[TestNumpyOp(inputs='x', outputs='x', mode="train", var=1), Delete(
```
TortoiseHam, line 52: I think the real test we are interested in here is Fuse((a->a,b), (del b)) and make sure that it runs without requiring b as an input to Fuse

aashokvardhan: Sure will add that test case too.

test/PR_test/integration_test/op/numpyop/meta/test_fuse.py
aashokvardhan: With the changes this test case is already failing but will add that assert too.

test/PR_test/integration_test/op/numpyop/meta/test_fuse.py
TortoiseHam: generally it's better if unit tests are black-box such that if someone later came and messed up the code without realizing why it had been written that way the test would catch it

test/PR_test/integration_test/op/numpyop/meta/test_fuse.py
TortoiseHam: the test case is failing?

aashokvardhan: I mean without the changes the unit test would fail.

TortoiseHam (approved):
LG

PR #1244, opened 2022-04-14
fix docstring

vbvg2008:


TortoiseHam (approved):
hopefully this doesn't break too much

PR #1252, opened 2022-04-26
Cutmix and Mixup reborn

DwijayDS:


fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -53,14 +53,13 @@ class CutMixBatch(TensorOp):
         AssertionError: If the provided inputs are invalid.
     """
     def __init__(self,
-                 inputs: str,
+                 inputs: Iterable[str],
```
TortoiseHam, line 56: The doc string for inputs and outputs needs to be updated. it would also be good to add an assert validating that exactly 2 keys are provided in each case

fastestimator/op/tensorop/augmentation/cutmix_batch.py
```diff
@@ -98,32 +97,42 @@ def _get_patch_coordinates(tensor: Tensor, x: Tensor, y: Tensor,
         """
         _, img_height, img_width = get_image_dims(tensor)
 
-        cut_x = img_width * x
-        cut_y = img_height * y
-        cut_w = img_width * tensor_sqrt(1 - lam)
-        cut_h = img_height * tensor_sqrt(1 - lam)
+        if tf.is_tensor(tensor):
```
TortoiseHam, line 100: is there any reason not to cast for torch also to reduce the if branching?

fastestimator/op/tensorop/augmentation/mixup_batch.py
```diff
@@ -50,30 +51,45 @@ def __init__(self,
                  mode: Union[None, str, Iterable[str]] = 'train',
                  ds_id: Union[None, str, Iterable[str]] = None,
                  alpha: float = 1.0,
-                 shared_beta: bool = True):
+                 shared_beta: bool = False):
         assert alpha > 0, "MixUp alpha value must be greater than zero"
         super().__init__(inputs=inputs, outputs=outputs, mode=mode, ds_id=ds_id)
```
TortoiseHam, line 56: adjust inputs/outputs doc string and init asserts

fastestimator/op/tensorop/augmentation/mixup_batch.py
```diff
@@ -50,30 +51,45 @@ def __init__(self,
...
         self.beta = None
         self.shared_beta = shared_beta
-        self.in_list, self.out_list = True, True
 
     def build(self, framework: str, device: Optional[torch.device] = None) -> None:
         if framework == 'tf':
             self.beta = tfp.distributions.Beta(self.alpha, self.alpha)
         elif framework == 'torch':
-            self.beta = torch.distributions.beta.Beta(self.alpha, self.alpha)
+            device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
```
TortoiseHam, line 65: device is given to you by the build function, no need to re-compute it here

fastestimator/backend/_get_shape.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import TypeVar
+
+import numpy as np
+import tensorflow as tf
+import torch
+
+Tensor = TypeVar('Tensor', tf.Tensor, torch.Tensor, np.ndarray)
+
+
+def get_shape(tensor: Tensor) -> Tensor:
```
TortoiseHam, line 24: init file should be updated with new function

fastestimator/op/tensorop/augmentation/cutmix_batch.py
DwijayDS: Initially we faced issues in tf, for which we included casting operation. Then while testing the updated code with pytorch, there was a multiplication issue created due to casting (Multiplication of numpy array with torch tensor). SInce casting wasnt necessary for torch, I just converted to if branch

TortoiseHam (approved):
LG

PR #1259, opened 2022-04-28
update jenkins

aashokvardhan:


CICD/PR_test/Jenkinsfile
```diff
@@ -61,6 +61,11 @@ pipeline {
                         }
                         finally {
                             sh 'docker rmi "$IMAGE_TAG"'
+                            def PR_IMAGE_ID = sh(script: 'docker images --filter "reference=*pr*" --format "{{.ID}}"', returnStdout: true).trim() as String
+                            if(PR_IMAGE_ID !=  "")
```
TortoiseHam, line 65: if we are deleting these images every time they are detected, why bother detecting them at all? can't we just issue the rmi command blindly?

aashokvardhan: So docker rmi would be throwing an error if we run it without any image ids to delete, to avoid this error we are verifying whether there are any stale images with 'pr' suffix in its name and than deleting them.

TortoiseHam (approved):
LG

PR #1261, opened 2022-05-03
Eval logging

aashokvardhan:


fastestimator/dataset/dataloader.py
```diff
@@ -156,6 +163,13 @@ def _get_fe_iterator(self):
                 return _MPPostBatchIter(self)
             return _MPPreBatchIter(self)
 
+    def __len__(self):
+        return self.fe_samples_to_yield
+
+    @property
+    def batch_size(self):
```
TortoiseHam, line 170: this might not work properly since I think the underlying data loader class has it's own batch_size attribute which we don't want to overwrite

TortoiseHam: we'll need to verify that though, not 100% confident

test/PR_test/integration_test/trace/test_logger.py
```diff
@@ -43,46 +42,44 @@ def test_on_begin_global_step(self):
...
+                             data=self.data, msg=self.on_batch_end_msg)
 
     def test_on_epoch_end_mode_train(self):
         logger = Logger()
         logger.system = sample_system_object()
         logger.system.global_step = 2
         logger.system.log_steps = 3
-        self._test_print_msg(func=logger.on_epoch_end, data=self.data, msg=self.on_epoch_end_train_msg)
-
-    def test_on_epoch_end_mode_eval(self):
```
aashokvardhan, line 67: Removed this test case, since we will not print the "FastEstimator-Eval: step:" since we are printing the progress.

TortoiseHam: why not update for the new print rather than just delete the test case?

fastestimator/estimator.py
```diff
@@ -39,10 +38,11 @@
 from fastestimator.trace.io.traceability import Traceability
 from fastestimator.trace.trace import EvalEssential, Logger, PerDSTrace, TestEssential, Trace, TrainEssential, \
     sort_traces
+from fastestimator.util.base_util import NonContext, Suppressor, to_list, to_set
 from fastestimator.util.data import Data, FilteredData
 from fastestimator.util.traceability_util import traceable
 from fastestimator.util.util import draw
-from fastestimator.util.base_util import to_set, to_list, NonContext, Suppressor
+from tensorflow.python.distribute.input_lib import DistributedDataset
```
TortoiseHam, line 45: this import sorting isn't right, external modules should appear before fe imports

fastestimator/estimator.py
```diff
@@ -155,17 +157,21 @@ def _prepare_traces(self, run_modes: Set[str]) -> None:
...
-            self.traces_in_use.insert(0, TrainEssential(monitor_names=self.monitor_names.union(extra_monitor_keys)))
+            self.traces_in_use.insert(0, TrainEssential(
+                monitor_names=self.monitor_names.union(extra_monitor_keys)))
             no_save_warning = True
             for trace in get_current_items(self.traces_in_use, run_modes=run_modes):
                 if isinstance(trace, (ModelSaver, BestModelSaver)):
                     no_save_warning = False
             if no_save_warning:
-                print("FastEstimator-Warn: No ModelSaver Trace detected. Models will not be saved.")
+                print(
```
TortoiseHam, line 167: it looks like the formatter settings may not be configured correctly. there's no reason for most of these format changes

fastestimator/estimator.py
```diff
@@ -339,34 +362,48 @@ def _run_epoch(self, eager: bool) -> None:
                                ds_id=self.system.ds_id,
                                steps_per_epoch=self.system.steps_per_epoch,
                                output_keys=trace_input_keys - network_output_keys | network_input_keys) as loader:
+
+                self.system.eval_log_steps = []
```
TortoiseHam, line 366: if we're adding a field to the system object, that should be done within the system object itself rather than by the estimator. Creating a new member variable here and then referencing it elsewhere will be confusing for future development and IDE linting won't work well. Also, if this variable is really required, then it should also probably be given a slightly different name, for example progress_log_steps or something to make it more clear what it's for

fastestimator/trace/trace.py
```diff
@@ -271,30 +300,49 @@ class Logger(Trace):
...
 
     def on_batch_end(self, data: Data) -> None:
         if self.system.mode == "train" and self.system.log_steps and (self.system.global_step % self.system.log_steps
                                                                       == 0 or self.system.global_step == 1):
-            self._print_message("FastEstimator-Train: step: {}; ".format(self.system.global_step), data)
+            self._print_message(
+                "FastEstimator-Train: step: {}; ".format(self.system.global_step), data)
+
+        if self.system.mode == "eval" and self.eval_step in self.system.eval_log_steps:
+            self._print_message(
```
TortoiseHam, line 329: this will break the current log visualization code

fastestimator/estimator.py
```diff
@@ -82,11 +83,14 @@ def __init__(self,
                  epochs: int,
                  train_steps_per_epoch: Optional[int] = None,
                  eval_steps_per_epoch: Optional[int] = None,
-                 traces: Union[None, Trace, Scheduler[Trace], Iterable[Union[Trace, Scheduler[Trace]]]] = None,
+                 traces: Union[None, Trace, Scheduler[Trace],
+                               Iterable[Union[Trace, Scheduler[Trace]]]] = None,
                  log_steps: Optional[int] = 100,
+                 eval_log_steps: List = [],
```
TortoiseHam, line 89: default args in a function call shouldn't be mutable. This would need to be an empty tuple instead. Also, typing should indicate what kind of values are intended here (List[int] for example.).

fastestimator/estimator.py
```diff
@@ -69,6 +69,7 @@ class Estimator:
         traces: What Traces to run during training. If None, only the system's default Traces will be included.
         log_steps: Frequency (in steps) for printing log messages. 0 to disable all step-based printing (though epoch
             information will still print). None to completely disable printing.
+        eval_log_steps:
```
TortoiseHam, line 72: include a description of what the field is, as with all other fields

fastestimator/summary/system.py
```diff
@@ -158,7 +163,8 @@ def _initialize_state(self) -> None:
         self.global_step = None
         self.epoch_idx = 0
         # Get a 64 bit random id related to current time
-        self.exp_id = int.from_bytes(uuid.uuid1().bytes, byteorder='big', signed=True) >> 64
+        self.exp_id = int.from_bytes(
```
TortoiseHam, line 166: formatter is not correct. this looks like your IDE line-width is set too low

fastestimator/estimator.py
```diff
@@ -97,6 +99,7 @@ def __init__(self,
                              total_epochs=epochs,
                              train_steps_per_epoch=train_steps_per_epoch,
                              eval_steps_per_epoch=eval_steps_per_epoch,
+                             eval_log_steps=eval_log_steps,
```
TortoiseHam, line 102: where does this variable get used within the estimator?

fastestimator/summary/system.py
```diff
@@ -259,7 +263,8 @@ def save_state(self, save_dir: str) -> None:
             'ds': {
                 mode: {key: value.__getstate__()
                        for key, value in ds.items() if hasattr(value, '__getstate__')}
-                for mode, ds in self.pipeline.data.items()
+                for mode,
```
TortoiseHam, line 266: formatting still?

fastestimator/trace/trace.py
```diff
@@ -214,25 +215,43 @@ class EvalEssential(Trace):
...
+        self.step_start = None
         self.eval_results = defaultdict(lambda: defaultdict(list))
 
     def on_epoch_begin(self, data: Data) -> None:
         self.eval_results = defaultdict(lambda: defaultdict(list))
+        self.eval_step = 0
+        self.elapse_times = []
+
+        if self.system.log_steps:
+            self.step_start = time.perf_counter()
```
TortoiseHam, line 232: this, as well as resetting self.eval_step to 0 should be done on_epoch_begin so that perf counting in higher epochs isn't messed up initially

fastestimator/dataset/dataloader.py
```diff
@@ -156,6 +155,12 @@ def _get_fe_iterator(self):
                 return _MPPostBatchIter(self)
             return _MPPreBatchIter(self)
 
+    def __len__(self):
+        return self.fe_samples_to_yield
+
+    def get_batch_size(self):
```
TortoiseHam, line 161:
```suggestion
    def get_batch_size(self) -> int:
```

fastestimator/trace/trace.py
```diff
@@ -214,25 +215,41 @@ class EvalEssential(Trace):
...
+
     def on_epoch_end(self, data: Data) -> None:
         for key, ds_vals in self.eval_results.items():
             for ds_id, vals in ds_vals.items():
                 if ds_id != '':
                     d = DSData(ds_id, data)
                     d.write_with_log(key, np.mean(np.array(vals), axis=0))
             data.write_with_log(key, np.mean(np.array([e for x in ds_vals.values() for e in x]), axis=0))
+        self.eval_step = 0
+        self.step_start = time.perf_counter()
```
TortoiseHam, line 252: put these on epoch begin and they can be taken out here

fastestimator/trace/trace.py
```diff
@@ -271,23 +289,36 @@ class Logger(Trace):
...
+            if self.eval_step in self.system.eval_log_steps:
+                self._print_message("Eval Progress: {}/{}; ".format(self.eval_step, self.system.eval_log_steps[-1]),
+                                    data)
+
     def on_epoch_end(self, data: Data) -> None:
         if self.system.mode == "train":
             self._print_message("FastEstimator-Train: step: {}; ".format(self.system.global_step), data, True)
-        elif self.system.mode == "eval":
+        elif self.system.mode == 'eval':
+            self.eval_step = 0
```
TortoiseHam, line 321: why reset eval_step here when it's getting reset on epoch_begin already?

fastestimator/dataset/dataloader.py
```diff
@@ -111,15 +111,14 @@ def __init__(self,
...
-                         batch_size=batch_size,
-                         sampler=sampler,
-                         num_workers=num_workers,
-                         persistent_workers=False,
-                         collate_fn=functools.partial(_pre_collate,
-                                                      try_fn=self.fe_collate_fn,
-                                                      postprocess_fn=postprocess_fn),
-                         worker_init_fn=lambda _: np.random.seed(random.randint(0, 2 ** 32 - 1)))
+        super().__init__(
+            dataset=dataset,
```
TortoiseHam, line 115: I don't think this formatting adjustment was necessary, now that you've got the IDE configured can you roll back these lines and see if it still changes their formatting?

fastestimator/estimator.py
TortoiseHam: in particular, why are we allowing the user to pass this value here when line 349 unilaterally overwrites the value stored in the system object?

fastestimator/dataset/dataloader.py
aashokvardhan: it still changes the format. i think its trying to keep the collate_fn in single line instead of 3.

fastestimator/summary/system.py
aashokvardhan: I updated the style but it still formats the same.

TortoiseHam (approved):
LG

PR #1271, opened 2022-05-16
updating jenkins

aashokvardhan:


CICD/PR_test/Jenkinsfile
```diff
@@ -61,10 +61,16 @@ pipeline {
...
+                            //To avoid throwing error when multiple pr are running in parallel
+                            try {
+                                def PR_IMAGE_ID = sh(script: 'docker images --filter "reference=*pr*" --format "{{.ID}}"', returnStdout: true).trim() as String
+                                if(PR_IMAGE_ID !=  "")
+                                {
+                                    sh 'docker rmi $(docker images --filter "reference=*pr*" -q)'
+                                }
+                            } catch (Exception e) {
+                                echo 'Exception occurred: ' + e.toString()
+                                sh 'Handle the exception!'
```
TortoiseHam, line 73: so if it breaks we just print 'handle the exception' and move on?

aashokvardhan: The command was added to remove stale images. But if there multiple PRs are running it would be throwing an error, since it would trying to delete a image which is running a container. To avoid deleting any images are currently used by current running container, adding this try catch block.

TortoiseHam (approved):

PR #1272, opened 2022-05-16
Include Tf installation Info

DwijayDS:


TortoiseHam:
the codacy issues look like they might be reasonable. what do you think?

PR #1277, opened 2022-05-25
fix issue with bool function

aashokvardhan:


fastestimator/backend/_focal_loss.py
```diff
@@ -64,64 +111,36 @@ def focal_loss(y_true: Tensor, y_pred: Tensor, gamma: float = 2.0, alpha: float
...
-            "Value of gamma should be greater than or equal to zero.")
-
-    if alpha:
-        if alpha != -1 and (alpha < 0 or alpha > 1):
-            raise ValueError(
-                "Value of alpha can either be -1 or None or within range (0, 1)")
-
-    if alpha and alpha != -1:
-        alpha = cast(alpha, y_pred)
+    if gamma is None and gamma < 0:
```
vbvg2008, line 114: gamma can never be None and <0 at the same time.

fastestimator/backend/_focal_loss.py
```diff
@@ -64,64 +111,36 @@ def focal_loss(y_true: Tensor, y_pred: Tensor, gamma: float = 2.0, alpha: float
...
+        raise ValueError("Value of gamma should be greater than or equal to zero.")
 
-    if gamma:
-        gamma = cast(gamma, y_pred)
+    if alpha is None and (alpha < 0 or alpha > 1):
```
vbvg2008, line 117: Same as above.

fastestimator/backend/_focal_loss.py
```diff
@@ -53,7 +99,8 @@ def focal_loss(y_true: Tensor, y_pred: Tensor, gamma: float = 2.0, alpha: float
                 positive vs negative examples or (-1/None) to ignore. Default = 0.25
         gamma: Exponent of the modulating factor (1 - p_t) to
                balance easy vs hard examples.
-        from_logits: Whether y_pred is logits (without softmax).
+        normalize: Whether to normalize along the batch size.
```
vbvg2008, line 102: "y_true: Ground truth class labels which should take values of 1 or -1."

I believe y_true should be 0 or 1 right? also we need to let user know the shape too

vbvg2008, line 102: Be more specific about what this normalization is doing.   Otherwise people don't know what's the difference between normalization and reduction.  

From the code it seems like the normalize is on a per-sample basis

fastestimator/backend/_focal_loss.py
```diff
@@ -47,13 +93,14 @@ def focal_loss(y_true: Tensor, y_pred: Tensor, gamma: float = 2.0, alpha: float
...
     Args:
-        y_true: Ground truth class labels which should take values of 1 or -1.
+        y_true: Ground truth class labels which should take values of 1 or 0.
         y_pred: Prediction score for each class, with a shape like y_true. dtype: float32 or float16.
         alpha: Weighting factor in range (0,1) to balance
                 positive vs negative examples or (-1/None) to ignore. Default = 0.25
         gamma: Exponent of the modulating factor (1 - p_t) to
                balance easy vs hard examples.
-        from_logits: Whether y_pred is logits (without softmax).
+        normalize: Whether to normalize focal loss along samples based on number of positive anchors per samples.
```
vbvg2008, line 102: let's not use the word `anchors`, as it is only specific to object detection.  Focal loss is more general than that

fastestimator/backend/_focal_loss.py
```diff
@@ -64,64 +111,36 @@ def focal_loss(y_true: Tensor, y_pred: Tensor, gamma: float = 2.0, alpha: float
...
-            ce = torch.nn.BCELoss(reduction="none")(
-                input=y_pred, target=y_true)
-            pred_prob = y_pred
-        ce = ce.view(ce.shape[0], -1)
+        focal_loss = pytorch_focal_loss(y_pred=y_pred, y_true=y_true, alpha=alpha, gamma=gamma, from_logits=from_logits)
     else:
         raise ValueError("Unsupported tensor type.")
 
-    p_t = (y_true * pred_prob) + ((1 - y_true) * (1 - pred_prob))
+    # normalize along the batch size based on number of positive anchors
```
vbvg2008, line 132: same

vbvg2008 (approved):

PR #1285, opened 2022-06-01
Fix in Advanced Tutorial 15 -> PyTorch workflow -> Start Finetuning

nikhilnanda21:
Updated code to freeze backbone under PyTorch workflow of Advanced Tutorial 15. Did not re-run the cell


TortoiseHam (approved):
LG

PR #1287, opened 2022-06-02
update fastcifar notebook

vbvg2008:
Getting rid of the extremely repetitive warning printing for this apphub (and only this apphub has the issue).


TortoiseHam (approved):
LG

PR #1288, opened 2022-06-03
update readme

vbvg2008:
Update readme's tensorflow version.

Also make Windows installation guide more general so it can be used for future TensorFlow installation.


TortoiseHam (approved):
LG

PR #1296, opened 2022-06-18
enabel warmup, eager, summary argument in run cli

vbvg2008:


vbvg2008:
also updated the test for FilterOp. In the test before current change, there is a small probability that it would fail (when the batch's idx is (18, 19, 20, 21, 22).   Changing the threshold to 19 will completely fix this issue.

By the way, the probability for the failure is 1 / 23_choose_5  = 1/33649, yet it happened :)

TortoiseHam (approved):
LG. this makes that particular pipeline test mostly pointless, but I guess if the failure actually occurred then worth handling

PR #1301, opened 2022-07-21
updating focal loss

aashokvardhan:


fastestimator/backend/_focal_loss.py
```diff
@@ -127,15 +133,19 @@ def focal_loss(y_true: Tensor,
     else:
         raise ValueError("Unsupported tensor type.")
 
+    focal_reduce_axis = [*range(1, len(focal_loss.shape))]
```
TortoiseHam, line 136: Have you tested this with TF in actual training (graph mode)? I suspect the shape call here won't work as expected

aashokvardhan: update to directly using tf.shape, and tested on unet code and it works both on tf and torch.

TortoiseHam (approved):
LG

PR #1307, opened 2022-08-18
Updated tensorflow functionality for Repeat Op

DwijayDS:


fastestimator/op/tensorop/meta/repeat.py
```diff
@@ -27,7 +27,6 @@
 Model = TypeVar('Model', tf.keras.Model, torch.nn.Module)
 
 
-@traceable()
```
TortoiseHam, line 30: this needs to go back

fastestimator/op/tensorop/meta/repeat.py
```diff
@@ -95,17 +97,17 @@ def __getstate__(self) -> Dict[str, List[Dict[Any, Any]]]:
     def forward(self, data: List[Tensor], state: Dict[str, Any]) -> List[Tensor]:
         # Set retain to true since might loop over a gradient aware op
         self.op.fe_retain_graph(True)
-
         data = {key: elem for key, elem in zip(self.inputs, data)}
         if isinstance(self.repeat, int):
-            for i in range(self.repeat - 1):
+            for i in self.for_range:
```
TortoiseHam, line 102:
```suggestion
            for _ in self.for_range:
```

fastestimator/op/tensorop/meta/repeat.py
```diff
@@ -161,10 +234,14 @@ def _tf_cond(self, cnd: List[Tensor], data: Dict[str, Tensor]) -> bool:
         Returns:
             Whether to continue looping.
         """
+        if isinstance(self.repeat, int):
+            # In this case we have 2 Forward calls for tf (one before and one after the while loop
+            # (For accurate Tf while loop functioning))
+            return tf.less(cnd, self.repeat - 1)
```
TortoiseHam, line 240: wouldn't this be self.repeat - 2?

DwijayDS: In line 190 , we start with cnd = tf.Constant(1), thats why 'self.repeat - 1' works

fastestimator/op/tensorop/meta/repeat.py
```diff
@@ -30,6 +30,25 @@
...
+        ops=[
+            LambdaOp(fn=lambda: 0, outputs="z"),
+            Repeat(AddOne(inputs="z", outputs="z"), repeat=lambda z: z < 6.5)
+            ]
+
+        Note : Here the argument ('z') of the lambda function used as repeat callable function is the key used by the
+               ops passed to the Repeat Op.
 
     Args:
         op: A TensorOp to be run one or more times in a row.
```
TortoiseHam, line 54: In line 58 we have that "If a function is provided, any TF ops which are wrapped by Repeat will not have access to the gradient tape, nor to previously deferred model update functions."

Is that still true? Does it now also apply to repeating N number of times?

TortoiseHam (approved):
LG

PR #1308, opened 2022-08-29
increase timeout period in nightly build

Vivek305:


aashokvardhan:
7 Hours looks like a long time, can we increase the instance type or number of instances to reduce the time ? Or can we run the apphub and fast estimator tests on separate instances to reduce time?

vbvg2008 (approved):

PR #1309, opened 2022-08-30
 Traceability Fix + Enhancement

TortoiseHam:
fix a performance issue where objects instantiated using input lists containing thousands of arguments build very slowly due to traceability overhead. Also fix an href issue with models and make numpy math literals display properly.


vbvg2008 (approved):
LGTM

PR #1312, opened 2022-08-31
making replacement optional in mscoco dataset, adding output channel arg in unet

vbvg2008:
To be consistent with other literatures that use MSCOCO, sometimes we need to filter out certain samples without replacement. This replacement feature enables this behavior.

The unet architecture has output channel to be hardcoded at 1, making it configurable will improve flexibility.

### DIFF
--- a/fastestimator/architecture/tensorflow/unet.py
+++ b/fastestimator/architecture/tensorflow/unet.py
@@ -20,11 +20,12 @@ from tensorflow.keras.models import Model
 
 
 # noinspection PyPep8Naming
-def UNet(input_size: Tuple[int, int, int] = (128, 128, 1)) -> tf.keras.Model:
+def UNet(input_size: Tuple[int, int, int] = (128, 128, 1), output_channel: int = 1) -> tf.keras.Model:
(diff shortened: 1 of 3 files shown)


fastestimator/architecture/tensorflow/unet.py
```diff
@@ -77,7 +78,7 @@ def UNet(input_size: Tuple[int, int, int] = (128, 128, 1)) -> tf.keras.Model:
     merge9 = concatenate([conv1, up9], axis=-1)
     conv9 = Conv2D(64, 3, **conv_config)(merge9)
     conv9 = Conv2D(64, 3, **conv_config)(conv9)
-    conv10 = Conv2D(1, 1, activation='sigmoid')(conv9)
+    conv10 = Conv2D(output_channel, 1, activation='sigmoid')(conv9)
```
TortoiseHam, line 81: doesn't TF do channel-last?

vbvg2008: Conv2D(output_channel, kernel_size, ...)

TortoiseHam (approved):
LG

PR #1322, opened 2022-10-05
updating ciFAIR data download

aashokvardhan:


vbvg2008:
need to update the test url too

vbvg2008 (approved):

PR #1323, opened 2022-10-05
one hot encoding

aashokvardhan:


fastestimator/op/numpyop/univariate/onehot.py
```diff
@@ -55,11 +55,14 @@ def forward(self, data: List[Union[int, np.ndarray]], state: Dict[str, Any]) ->
...
 
     def _apply_onehot(self, data: Union[int, np.ndarray]) -> np.ndarray:
-        class_index = np.array(data)
-        assert "int" in str(class_index.dtype)
-        assert class_index.size == 1, "data must have only one item"
-        class_index = class_index.item()
-        assert class_index < self.num_classes, "label value should be smaller than num_classes"
-        output = np.full((self.num_classes), fill_value=self.label_smoothing / self.num_classes, dtype="float32")
-        output[class_index] = 1.0 - self.label_smoothing + self.label_smoothing / self.num_classes
+        data = np.array(data).astype(np.uint8)
```
vbvg2008, line 58: uint8 can represent at most 256 classes.  It won't work on ImageNet (which as 1000 classes)

vbvg2008, line 58: we should assert the data that comes in is integer.

fastestimator/op/numpyop/univariate/onehot.py
```diff
@@ -55,11 +55,15 @@ def forward(self, data: List[Union[int, np.ndarray]], state: Dict[str, Any]) ->
...
     def _apply_onehot(self, data: Union[int, np.ndarray]) -> np.ndarray:
-        class_index = np.array(data)
-        assert "int" in str(class_index.dtype)
-        assert class_index.size == 1, "data must have only one item"
-        class_index = class_index.item()
-        assert class_index < self.num_classes, "label value should be smaller than num_classes"
-        output = np.full((self.num_classes), fill_value=self.label_smoothing / self.num_classes, dtype="float32")
-        output[class_index] = 1.0 - self.label_smoothing + self.label_smoothing / self.num_classes
+        data = np.array(data)
+        assert "int" in str(data.dtype)
```
vbvg2008, line 59: add an assert message saying the data type must be an integer

vbvg2008 (commented):
In the docstring, stating that label starts from 0.

fastestimator/op/numpyop/univariate/onehot.py
aashokvardhan: added assert message

vbvg2008 (approved):
LG

PR #1330, opened 2022-10-12
fix the cifar/cifairlabel format issues that failed nightly test

vbvg2008:
To support multi-dimensional cross entropy, the label shape can no longer be ambiguous like before.

For example, in torch's sparse cross entropy, we require the shape of `y_pred` to be in (B, ..., C) and `y_true`'s shape needs to be (B, ...) 

However, in case of Cifar/Cifair, y_true's shape is (B, ..., 1), which now violates the shape requirement. This violation used to be allowed under the 1-dimensional label assumption. But since the assumption no longer holds after the multi-dimensional cross entropy support, we need to start enforcing label's shape. 

Hence this PR.

### DIFF
--- a/fastestimator/dataset/data/cifair10.py
+++ b/fastestimator/dataset/data/cifair10.py
@@ -71,8 +71,7 @@ def load_data(root_dir: str = None, image_key: str = "x", label_key: str = "y") 
     fpath = os.path.join(image_extracted_path, 'test_batch')
     x_test, y_test = _load_batch(fpath)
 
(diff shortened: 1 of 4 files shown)


TortoiseHam:
If you're changing the dimensionality of the tensors won't the transpose functions below the modified lines also need to be changed?

vbvg2008:
transpose is applied on x only,    I'm only changing y

TortoiseHam (approved):
LG

PR #1351, opened 2022-11-24
fix restore wizard issue

vbvg2008:
during `fe.build`, when model's `optimizer_fn=None`(like when calling a non-trainable model),  `Restorewizard` will try to save its optimizer, which will lead to error.

This PR will fix this behavior by checking whether the optimizer exists while saving the state, during loading, it would check for presence of optimizer file and load accordingly.


TortoiseHam (approved):
LG

PR #1357, opened 2023-01-10
rplace  sklearn dependency with scikit-learn

vbvg2008:
`uncertainty-calibration` is using sklearn as dependency, but its latest release corrected the dependency, therefore we will use their latest release.


TortoiseHam (approved):
LG

PR #1374, opened 2023-04-25
R1.5

vbvg2008:


docker/stable/Dockerfile.gpu
```diff
@@ -34,7 +34,7 @@ RUN pip install --no-cache-dir \
     ipython \
     ipykernel \
     ipywidgets \
-    papermill \ 
+    papermill \
```
TortoiseHam, line 37: don't the stable docker files also need to change the fe version to specify 1.5.1?

vbvg2008: nice catch, updated.

TortoiseHam (approved):
LG

PR #1391, opened 2023-05-15
Inheriting PyTorch Concat Dataset in the combined dataset

Nishanksingla:
As the per the discussion with Michael, we have concluded that split function does not aligns with the CombinedDataset because if we do the split it will impact the original individual datasets and the only way to remove this drawback is to copy the data into the combined dataset as well which will increase the memory footprint.

Users can split the dataset before combining them. 

Note: We are inheriting from the PyTorch concatdataset for it's optimized implementation.


fastestimator/dataset/combined_dataset.py
```diff
@@ -14,70 +14,40 @@
...
+from torch.utils.data import ConcatDataset
 
 from fastestimator.dataset.dataset import FEDataset
 from fastestimator.util.traceability_util import traceable
 
 
 @traceable()
-class CombinedDataset(FEDataset):
-    def __init__(self, datasets: list) -> None:
+class CombinedDataset(ConcatDataset, FEDataset):
```
TortoiseHam, line 26: In this case it's no longer an FEDataset. It can just inherit the ConcatDataset. corresponding changes in the unit test text

Nishanksingla: Then we can accept list of dataset instead of list of FEDataset.

fastestimator/dataset/data/usps.py
```diff
@@ -112,18 +113,16 @@ def load_data(root_dir: Optional[str] = None) -> Tuple[LabeledDirDataset, Labele
     if not os.path.exists(train_base_path):
         if not os.path.exists(train_compressed_path):
             print("Downloading train data to {}".format(root_dir))
-            wget.download('http://statweb.stanford.edu/~tibs/ElemStatLearn/datasets/zip.train.gz',
-                          root_dir,
-                          bar=bar_custom)
+            _download_file_from_google_drive("1NlaCnlhV-PA_Rek8w8eQqJUeYCZJ0olG", train_compressed_path)
+
```
vbvg2008, line 117: can you remove the additional new line according to codacy comment?

fastestimator/dataset/data/usps.py
```diff
@@ -112,18 +113,16 @@ def load_data(root_dir: Optional[str] = None) -> Tuple[LabeledDirDataset, Labele
...
         _write_data(train_images, train_labels, train_base_path, "train")
 
     if not os.path.exists(test_base_path):
         if not os.path.exists(test_compressed_path):
             print("Downloading test data to {}".format(root_dir))
-            wget.download('http://statweb.stanford.edu/~tibs/ElemStatLearn/datasets/zip.test.gz',
-                          root_dir,
-                          bar=bar_custom)
+            _download_file_from_google_drive("1lagM-V1nmAdS3Uz9Uk4bB9cKqE_agt59", test_compressed_path)
+
```
vbvg2008, line 125: same, we can remove this additional new line within the if block.

TortoiseHam (approved):
LG

PR #1413, opened 2023-06-27
updating google download retry

aashokvardhan:


fastestimator/util/google_download_util.py
```diff
@@ -66,3 +68,25 @@ def _download_file_from_google_drive(file_id: str, destination: str) -> None:
...
+    for _ in range(max_retries):
+        try:
+            time.sleep(2)
+            _download_file_from_google_drive(file_id, destination)
+            if os.path.exists(destination):
+                break
+        except Exception:
+            continue
+
+    print(f"Couldn't download {file_id} after {max_retries} retries.")
```
TortoiseHam, line 92: Should this raise an exception rather than just print since we're going to run into an exception later anyways?

fastestimator/util/google_download_util.py
```diff
@@ -66,3 +68,25 @@ def _download_file_from_google_drive(file_id: str, destination: str) -> None:
...
+    This method will try to download the file for given number of retries till successful.
+
+    Args:
+        file_id: File ID of Google drive URL.
+        destination: Destination path where the data needs to be stored.
+        max_retries: max number of retries.
+    """
+    for _ in range(max_retries):
+        try:
+            time.sleep(2)
```
TortoiseHam, line 85: can we sleep for a random interval between 5 and 10 seconds rather than hardcoded 2 in case what is happening here is a result of multi-threaded tests all firing requests simultaneously?

fastestimator/util/google_download_util.py
```diff
@@ -66,3 +68,25 @@ def _download_file_from_google_drive(file_id: str, destination: str) -> None:
...
+            _download_file_from_google_drive(file_id, destination)
```
TortoiseHam, line 86: this checks for file existence before attempting the download right?

aashokvardhan: yes if file exists it would break the look.

fastestimator/util/google_download_util.py
aashokvardhan: yes, we can raise a exception over here itself instead of raising it again later.

fastestimator/util/google_download_util.py
aashokvardhan: updated to sleep for 5-10 seconds

fastestimator/util/util.py
```diff
@@ -214,6 +215,36 @@ def _custom_tf_print(*args, **kwargs):
...
 
 
+def validate_file(file_path: str) -> bool:
+    """Validate whether file valid or not.
+
+    Args:
+        file_path (str): location of the input file.
+
+    Returns:
+        bool: Whether the is valid.
```
TortoiseHam, line 225:
```suggestion
        Whether the file is valid.
```

fastestimator/util/util.py
```diff
@@ -214,6 +215,36 @@ def _custom_tf_print(*args, **kwargs):
     print_v2(*args, **kwargs)
 
 
+def validate_file(file_path: str) -> bool:
+    """Validate whether file valid or not.
+
+    Args:
+        file_path (str): location of the input file.
```
TortoiseHam, line 222:
```suggestion
        file_path: location of the input file.
```

fastestimator/util/util.py
```diff
@@ -214,6 +215,36 @@ def _custom_tf_print(*args, **kwargs):
     print_v2(*args, **kwargs)
 
 
+def validate_file(file_path: str) -> bool:
+    """Validate whether file valid or not.
```
TortoiseHam, line 219:
```suggestion
    """Validate whether file is valid or not.
```

fastestimator/util/util.py
```diff
@@ -214,6 +215,36 @@ def _custom_tf_print(*args, **kwargs):
...
+                with tarfile.open(file_path) as img_tar:
+                    _ = img_tar.getmembers()
+            else:
+                import gzip
+                f = gzip.open(file_path, 'rb')
+                _ = f.read()
+        elif not (os.path.exists(file_path) and os.path.getsize(file_path) > 0):
+            return False
+        return True
+    except Exception as e:
```
TortoiseHam, line 244: could drop the e or maybe do 'Exception as _' to avoid this?

TortoiseHam:
suppose that 2 test threads both kick off downloads simultaneously, then they will corrupt the zip file as the other one is trying to download it right? I think we might need to wrap the inner download function in a global lock to ensure we're only running one request to google drive at a time. Alternatively at the very least I think we'd need the loop to be [(validate) -> (random sleep) -> (validate) -> (attempt download)] rather than [(attempt download) -> (validate) -> (random sleep)]

fastestimator/dataset/data/cifair10.py
```diff
@@ -52,9 +52,8 @@ def load_data(root_dir: str = None, image_key: str = "x", label_key: str = "y")
     image_extracted_path = os.path.join(root_dir, 'ciFAIR-10')
 
     if not os.path.exists(image_extracted_path):
-        if not os.path.exists(image_compressed_path):
-            print("Downloading data to {}".format(root_dir))
-            _download_file_from_google_drive('1dqTgqMVvgx_FZNAC7TqzoA0hYX1ttOUq', image_compressed_path)
+        print("Downloading data to {}".format(root_dir))
```
TortoiseHam, line 55: why are we removing these checks for if we've already downloaded the zip? in that case we definitely should check for whether we already have a valid zip before re-downloading

fastestimator/util/util.py
```diff
@@ -214,6 +215,37 @@ def _custom_tf_print(*args, **kwargs):
...
+        elif Path(file_path).suffix == '.gz':
+            if file_path.endswith('.tar.gz'):
+                import tarfile
+                with tarfile.open(file_path) as img_tar:
+                    _ = img_tar.getmembers()
+            else:
+                import gzip
+                f = gzip.open(file_path, 'rb')
+                _ = f.read()
+        elif not (os.path.exists(file_path) and os.path.getsize(file_path) > 0):
```
TortoiseHam, line 241: why does the file size need to be >0 here? wouldn't an empty file that passed the above tests technically still be valid?

fastestimator/dataset/data/cifair10.py
aashokvardhan: Currently we are validating before download. Instead of doing it every time we are using google download, we are just moving the if check instead the download, it would check the file and download if the file doesn't exist.

fastestimator/util/util.py
aashokvardhan: Doing Exception as _ is raising codacy errors.

fastestimator/util/util.py
```diff
@@ -214,6 +215,38 @@ def _custom_tf_print(*args, **kwargs):
...
+                import tarfile
+                with tarfile.open(file_path) as img_tar:
+                    _ = img_tar.getmembers()
+            else:
+                import gzip
+                f = gzip.open(file_path, 'rb')
+                _ = f.read()
+        return True
+    except Exception as e:
+        print(e)
```
TortoiseHam, line 246: printing e here seems pretty messy though? should we just `except Exception` and ignore the codacy complaint?

aashokvardhan: But would deleting the file cause issues with different threads downloading it?

TortoiseHam (approved):
LG

PR #1422, opened 2023-07-17
introduced a fail safe for search visualization

DwijayDS:
1. Now we throw error if no data is available to visualize
2. Numpy array is supported in search data parsing

### DIFF
--- a/fastestimator/search/visualize/vis_util.py
+++ b/fastestimator/search/visualize/vis_util.py
@@ -23,6 +23,8 @@ from fastestimator.search.search import Search
 from fastestimator.summary.summary import ValWithError
 from fastestimator.util.base_util import to_set
 
+import numpy as np
+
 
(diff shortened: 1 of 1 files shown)


fastestimator/search/visualize/vis_util.py
```diff
@@ -60,6 +62,9 @@ def __init__(self, search: Search, ignore_keys: Union[None, str, Sequence[str]]
             self.params.discard(key)
             self.results.discard(key)
 
+        # Check if any results left to use
+        assert len(self.results) != 0, "No data found!!!!!"
```
TortoiseHam, line 66:
```suggestion
        assert len(self.results) != 0, f"No results found after ignoring keys: {ignore_keys}"
```

TortoiseHam (approved):
LG

PR #1432, opened 2023-08-10
adding new min and new max to minmax operator

vbvg2008:
# Requirements

`Minmax` should be able to output results in arbitrary min and max value defined by user. 

# Target Audience

* Users

Adding new arguments: `new_min` and `new_max` to `Minmax`, such that the results will be further scaled by the `new_min` and `new_max`. 

# Known Issues / Limitations

Nope, not at all 😉 . 

Except when AGI is ready it would obsolete everything, including this PR.

### DIFF
--- a/fastestimator/op/numpyop/univariate/minmax.py
+++ b/fastestimator/op/numpyop/univariate/minmax.py
@@ -49,6 +49,7 @@ class Minmax(NumpyOp):
         self.new_min = new_min
         self.new_max = new_max
         self.in_list, self.out_list = True, True
+        assert new_max > new_min, "the new_max should be greater than new_min."
 
     def forward(self, data: List[np.ndarray], state: Dict[str, Any]) -> List[np.ndarray]:
         return [self._apply_minmax(elem) for elem in data]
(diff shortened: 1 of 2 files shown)


fastestimator/op/numpyop/univariate/minmax.py
```diff
@@ -33,15 +33,21 @@ class Minmax(NumpyOp):
...
                  outputs: Union[str, Iterable[str]],
                  mode: Union[None, str, Iterable[str]] = None,
                  ds_id: Union[None, str, Iterable[str]] = None,
-                 epsilon: float = 1e-7):
+                 epsilon: float = 1e-7,
+                 new_min: float = 0.0,
+                 new_max: float = 1.0):
         super().__init__(inputs=inputs, outputs=outputs, mode=mode, ds_id=ds_id)
         self.epsilon = epsilon
+        self.new_min = new_min
```
TortoiseHam, line 49: should we have an assert for new_max > new_min?

vbvg2008: Added, good point.

PR #1443, opened 2023-10-26
Lora apphub

vbvg2008:
# Requirements

Need to stay trendy by having some lora examples in apphub

# Target Audience

* Users
* Developers

implemented tf, torch, notebook of LoRA implementation in FE. 

# Known Issues / Limitations

I don't think so.

## Summary by CodeRabbit

New Features:
- Introduced the "Foundation Model" section in the documentation, providing resources on the "Low-Rank Adaptation" model.
- Added TensorFlow and PyTorch implementations of the LoRA (Learnable ReLU Activation) training for the Vision Transformer (ViT) architecture, enhancing the model's performance.

Bug Fix:
- Corrected the error handling in the `fastestimator/backend/_load_model.py` file, improving the software's reliability.

Chore:
- Added bash scripts to facilitate running FastEstimator training scripts and Jupyter notebooks, simplifying the process for users.

Documentation:
- Updated the FastEstimator library codebase, including the addition of new classes and functions, providing more detailed and comprehensive information for users.


apphub/foundation_model/lora/lora_tf.py
```diff
@@ -0,0 +1,299 @@
...
+        loaded_vars = np.load(save_path)
+        for var_name, var in lora_vars.items():
+            var.assign(loaded_vars[var_name])
+
+    def save_lora_weights(self, save_dir, name="lora_weight"):
+        lora_vars = {var.name: var.numpy() for var in self.trainable_variables if "lora_w" in var.name}
+        save_path = os.path.join(save_dir, "{}.npz".format(name))
+        print("Saving LoRA weights to {}...".format(save_path))
+        np.savez(save_path, **lora_vars)
+        return save_path
```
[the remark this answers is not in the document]
vbvg2008: oh that's nice! shut up!

fastestimator/backend/_load_model.py
```diff
@@ -49,7 +49,6 @@ def load_model(model: Union[tf.keras.Model, torch.nn.Module], weights_path: str,
     Raises:
         ValueError: If `model` is an unacceptable data type.
     """
-    assert hasattr(model, "fe_compiled") and model.fe_compiled, "model must be built by fe.build"
```
TortoiseHam, line 52: ???

apphub/foundation_model/lora/lora_torch.py
```diff
@@ -0,0 +1,570 @@
...
+            nn.init.zeros_(w_B.weight)
+
+    def forward(self, x):
+        out = self.base_encoder(x)
+        return out
+
+    def load_lora_weights(self, save_path):
+        print("Loading LoRA weights from {}...".format(save_path))
+        self.lora_modules.load_state_dict(
+            torch.load(save_path, map_location='cpu' if torch.cuda.device_count() == 0 else None))
```
TortoiseHam, line 392: use our util gpu detection method instead of device_count() for compatibility with Mac

fastestimator/backend/_load_model.py
vbvg2008: I realize our load model can be used for non-fe compiled model for convenient loading too, such as loading SAM encoder weight in the torch implementation. Removing the assertiton because of that.

TortoiseHam: I feel like we're opening a can of worms for majority use case though. wouldn't someone who just wants to load SAM simply use model.load_state_dict() ?

vbvg2008: `model.load_state_dict()` won't work for multi-gpu system.   also there's an additional step of `preprocess_torch_weights` for pytorch weights too if a model is saved in mutli-gpu system but loaded in single -gpu system.  

That's why I need to call our own backend function rather than duplicating the code.

TortoiseHam: hmm, what if you just add in fe_compiled manually to it?

vbvg2008: I think removing the assertion is better, because it can at least allow other pytorch models to leverage our `load_model` capability.

TortoiseHam (approved):
LG

PR #1448, opened 2023-11-06
update sliding slicer

aashokvardhan:
# Requirements

1. slicer stop padding the first time we encounter it

# Target Audience

* Data Scientists

Requirement 1 was addressed by updating padding logic to take, window_size into consideration and breaking when the end is reached for the first time.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the slicing mechanism to prevent overshooting the target shape.
- **Tests**
	- Updated unit tests to reflect the changes in the slicing mechanism.


TortoiseHam (approved):
LG

PR #1478, opened 2024-05-22
upgrading pytorch to fix circular dependency

aashokvardhan:
# Requirements

1.upgrading torch to remove circular dependency issues.

## Summary by CodeRabbit

- **Documentation**
  - Updated installation instructions to use PyTorch 2.1.1 (previously 2.0.1) for both CPU and GPU.

- **Chores**
  - Updated PyTorch, TorchVision, and TorchAudio versions in Docker configurations for both CPU and GPU.
  - Updated version check for PyTorch in the codebase to ensure compatibility with PyTorch 2.1.1.
  - Updated `requests` library version in the setup configuration from 2.28.2 to 2.32.2.


DwijayDS (approved):
LG

PR #1508, opened 2025-04-08
Remove tensorflow backend

aashokvardhan:
Removing backend tensorflow dependency


fastestimator/backend/_binary_crossentropy.py
```diff
@@ -14,13 +14,12 @@
 # ==============================================================================
 from typing import Dict, Optional, TypeVar
 
-import tensorflow as tf
 import torch
 
 from fastestimator.backend._reduce_mean import reduce_mean
 
-Tensor = TypeVar('Tensor', tf.Tensor, torch.Tensor)
```
Vivek305, line 22: We could remove the `TypeVar` tensor here and just use `torch.Tensor` for typing

Vivek305 (changes requested):

fastestimator/backend/_get_gradient.py
```diff
@@ -67,10 +47,7 @@ def get_gradient(target: Tensor,
     Raises:
         ValueError: If `target` is an unacceptable data type.
     """
-    if tf.is_tensor(target):
-        with NonContext() if higher_order else tape.stop_recording():
-            gradients = tape.gradient(target, sources)
-    elif isinstance(target, torch.Tensor):
+    if isinstance(target, torch.Tensor):
```
DwijayDS, line 50: DO we need block that starts on line 58 now (`if isinstance(sources, torch.Tensor):`)? How would the functionality change if we remove that line?

aashokvardhan: I think it's better to leave it. Since in case we send numpy to tensor flow unexpectedly it would raise an error, When removed it would be trying torch implementation for tensorflow tensor.

Vivek305 (approved):
LG

PR #1513, opened 2025-11-18
fixing-test-cases

aashokvardhan:
# Requirements

1. Updating the test cases


fastestimator/dataset/data/mnist.py
```diff
@@ -39,14 +39,17 @@ def load_data(image_key: str = "x", label_key: str = "y", root_dir: str = None)
...
-    eval_data = datasets.MNIST(root_dir, train=False, transform=None)
+    mnist = fetch_openml('mnist_784', version=1, parser='auto')
+    # Access data and labels
+    X, y = mnist.data, mnist.target
+    x = X.to_numpy().reshape(-1, 28, 28)
+    y = y.to_numpy()
 
-    x_train = np.array([image for image, _ in train_data]).astype(np.float32)
-    y_train = np.array([clas for _, clas in train_data])
+    x_train = x[:60000].astype(np.float32)
```
chm123, line 48: why are we limiting this to 60k?

aashokvardhan: While the tensor flow and PyTorch datasets split the training and validation dataset, skimage provides it as single data frame combining training and validation where first 60k are training and 10k are validation.

vbvg2008:
might need to update the README too :)

